PR #70, opened 2020-10-06
Adx 281 Put geojson in the data store

jonathansberry:
This PR provide a bug fix and regression test to ensure that the format of geojson files are correctly guessed during the validation phase of creating a new resource.  Previously geojson files that didn't not have an explicitly set format were guessed to be of type JSON".

This PR should be merged after https://github.com/fjelltopp/ckanext-unaids/pull/69 as it requires the testing setup to be in place.


ckanext/unaids/tests/test_plugin.py
```diff
@@ -0,0 +1,49 @@
...
+import logging
+
+
+class TestPlugin(helpers.FunctionalTestBase):
+    '''Tests for the ckanext.example_iauthfunctions.plugin module.
+
+    Specifically tests that overriding parent auth functions will cause
+    child auth functions to use the overridden version.
+    '''
+    def setup(self):
```
tomeksabala, line 18: You don't need to use custom setup and teardown methods if using `helpers.FunctionalTestBase`. The parent class will clear and setup db for you. All you need to do is define which plugins you need with a field:
look at example in: ckanext-restricted/ckanext/restricted/tests/test_plugin.py
```
class TestRestrictedPlugin(helpers.FunctionalTestBase):

    _load_plugins = ['restricted']

    def test_restricted_resource_update(self):
(...)
```

tomeksabala (changes requested):

ckanext/unaids/tests/test_plugin.py
```diff
@@ -0,0 +1,49 @@
...
+        '''
+        # We have to unload the plugin we loaded, so it doesn't affect any
+        # tests that run after ours.
+        ckan.plugins.unload('unaids')
+
+    def test_geojson_format_guessed_correctly(self):
+        '''
+        Test that the format of a geojson file is guessed correctly.
+        '''
+        logging.warning('RUNNING Test')
```
tomeksabala, line 38: This should be removed. Or if it needs to be here make it `debug` level.

PR #72, opened 2020-10-13
Fixed the order we import CSS files

Manoj-nathwani:
# Problem
- We lost our custom UNAIDS styles when upgrading to ckan 2.9

image

# Solution
- The order in which the base css file has moved lower down in the htm, so we need to shuffle our custom css imports lower down in the order to overwrite the base css
- This was technically always an issue which wasn't causing any ugly effects, so it went unnoticed 😅  

image

### DIFF
--- a/ckanext/unaids/presets.json
+++ b/ckanext/unaids/presets.json
@@ -28,6 +28,33 @@
         }
     	}
     },{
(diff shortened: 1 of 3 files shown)


tomeksabala (approved):
Nice catch. LGTM.

PR #74, opened 2020-10-14
migrated scripts to webassets

Manoj-nathwani:
# Problem
- As per the [python3 migrate guide](https://github.com/ckan/ckan/wiki/Python-3-migration-guide-for-extensions#web-assets-js-and-css), we need move over to using `webassets.yml` as it has a smarter way of grouping imports of css/js files
- I'll migrate over the scripts sitting in `/fanstatic` directory

# Solution
- Implemented `webassets.yml` and updated all references where we want to import the files


tomeksabala (approved):
LGTM

PR #75, opened 2020-10-15
ADX-417 Autogenerate Data Explorers for geojson files

jonathansberry:
The recline_view creates the "Data Explorer" in ckan that lets you browse, filter and graph data. 

However, changing which files get an autocreated recline view requires a hard code change.  I think i will open up a PR with core ckan to change this to a config change. 

In the meantime I have this pretty neat fix which involves creating a new unaids specific recline view from the same base as the orgiinal recline view.  This unaids recline view can then be edited as desired.  

As this is a plugin shipped by default with ckan core, it can't be overridden by the usual git clone/edit/install route, and the extending the reclineview base works well.

This is linked to https://github.com/fjelltopp/adx_develop/pull/18, and https://github.com/fjelltopp/adx_deploy/pull/52

### DIFF
--- a/ckanext/unaids/plugin.py
+++ b/ckanext/unaids/plugin.py
@@ -15,6 +15,7 @@ from ckanext.unaids.helpers import (
     get_user_obj
 )
 import ckan.plugins.toolkit as toolkit
+from ckanext.reclineview.plugin import ReclineViewBase
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


ckanext/unaids/plugin.py
```diff
@@ -99,3 +100,32 @@ def get_validators(self):
...
+
+
+class UNAIDSReclineView(ReclineViewBase):
+    '''
+    This override of the recline view plugin allows data explorers to be auto
+    created for geojson files.
+    '''
+
+    def info(self):
+        return {'name': 'unaids_recline_view',
```
jonathansberry, line 112: We have to change this line so the view is uniquely identified and uncoupled fro the original recline view.

tomeksabala (approved):
Looks good. I'll merge it so it gets nightly deployed.

PR #80, opened 2020-10-30
ADX-463 Fix validation logic.

jonathansberry:
This uses the new IDataValidation interface to only validate data that has a schema field specified, and to trigger asyncronous whole-package validation if the "validate_package" field is present.


tomeksabala (approved):
LGTM

PR #81, opened 2020-11-03
ADX-284 Clean up the new collaborator form.

jonathansberry:
This just touches up a template for the new Collaborators feature.  The feature seems to work well and will be really valuable I think moving forwards.  

There is possible overlap with ckanext-restricted here.  However, adding a collboarator to the package under the role (member) is effectively shorthand for going through adding an individual user to every seperate resource. 

A possible hole in this feature is that it would be nice to add another org as a collaborator so that all users of the orgs can contribute to the dataset.  But something to consider for the mid to long-term future I think.  

Related to: https://github.com/fjelltopp/adx_develop/pull/24


tomeksabala (approved):
LGTM

PR #84, opened 2020-11-10
ADX-460-dataset-cloning

Manoj-nathwani:
# This feature
- [x] Validate a real organization_id has been saved
- [x] UI shows a banner asking the recipient to accept the dataset transfer
- [x] Backend code to make dataset transfer
- [x] Success flash banner on transfer complete 
- [x] Clear UX informing the sender to inform the recipient of the transfer request

# No longer in the scope of this task
- [ ] ~Validate recipient has view-access to the dataset before transfer~
- [ ] ~Email to recipient so they can action the invitation~
- [ ] ~Email to sender to let them know it's been migrated~


Manoj-nathwani:
# Emailing transfer request
- Looks like emailing a notification to the recipient may be pretty complex-
- I'm shelving this for now
- Posting the code here just incase we want to add it later

## Backend (untested)
```
from ckan.lib.base import render_jinja2
import ckan.lib.mailer as mailer

log = logging.getLogger(__name__)

try:
	email_body_context = {
	    'recpipient_name': 'aaaaaaaaa',
	    'recpipient_org_name': 'bbbbbb',
	    'sender_name': 'ccccccc',
	    'dataset_name': 'ddddd',
	    'dataset_url': 'eeeeee'
	}
	email_body = render_jinja2(
	    'emails/dataset_transfer_requested.txt',
	    email_body_context
	)
	mailer.mail_recipient(
	    'Manoj Nathwani',
	    'me@me.com',
	    _('You have been sent a dataset transfer'),
	    email_body
	)
except Exception as e:
	log.warning(('dataset_transfer_requested email: '
	             'Failed to send mail to "{0}": {1}').format(user_id, e))
```

## Email template (untested)
templates/emails/dataset_transfer_requested.txt
```
{% trans %}Dear{% endtrans %} {{ recpipient_name }},

{% trans %}{{ sender_name }}{% endtrans %} {% trans %}would like to transfer the{% endtrans %}{{ dataset_name }} dataset to your organization

{% trans %}Dear{% endtrans %} {{ recpipient_name }}, {% trans %}would like to transfer the{% endtrans %} {{ dataset_name }} {% trans %}to{% endtrans %} {{ recpipient_org_name }}.{% endtrans %}

{% trans %}Please action the transfer request here:{% endtrans %} {{ dataset_url }}
```

Manoj-nathwani:
image

added success flash on completion ✅

ckanext/unaids/blueprints/unaids_dataset_transfer.py
```diff
@@ -0,0 +1,47 @@
...
+    url_prefix=u'/validation'
+)
+
+
+def process_dataset_transfer(dataset_id):
+    dataset = toolkit.get_action('package_show')({}, {'id': dataset_id})
+    toolkit.check_access(
+        'organization_update',
+        {'id': dataset['org_to_allow_transfer_to']}
+    )
```
Manoj-nathwani, line 25: This is important to check

ckanext/unaids/theme/templates/package/read.html
```diff
@@ -3,6 +3,20 @@
 {% set pkg = c.pkg_dict %}
 
 {% block package_description %}
+
+  {% if pkg.org_to_allow_transfer_to %}
+    {% if h.check_access('organization_update', {'id': pkg.org_to_allow_transfer_to }) %}
```
Manoj-nathwani, line 8: This is important to check

Manoj-nathwani: This now needs to be changed to:
```
user_organizations = \
        logic.get_action('organization_list_for_user')\
        ({'user': c.user}, {})
    valid = dataset['org_to_allow_transfer_to'] in [
        str(org['id'])
        for org in user_organizations
        if org['capacity'] in ['admin', 'editor']
    ]
```

Manoj-nathwani: Moved this out into a helper 👍

Manoj-nathwani:
# After sending a transfer request UX...

# What members of the sender’s org will see:

image

# What editors/admins of the recipient org sees

image


# What everyone else sees

image

ckanext/unaids/blueprints/unaids_dataset_transfer.py
```diff
@@ -0,0 +1,54 @@
...
+    url_prefix=u'/validation'
+)
+
+
+def process_dataset_transfer(dataset_id):
+    dataset = toolkit.get_action('package_show')({}, {'id': dataset_id})
+    
+    # validate user can transfer dataset
+    valid = helpers.check_organization_update_access(
+        dataset['org_to_allow_transfer_to']
```
jonathansberry, line 23: Writing our own authorisation logic is fine, but let's register it properly as an auth check and call it with check access. I think you need to do the following (this is broadly correct, but may not be 100% correct): 
 1) Define a function with all your auth logic.  It should take arguments "context" and "data_dict", being dictionaries of the form `{'user': '<user_object>'}` and  `{'org_to_allow_transfer_to': '<org_id>'}` respectively.  The function should return a dict: `{'success': True}` if it passes. 
 2) Make the unaids plugin implement the [IAuthFunctions interface](https://docs.ckan.org/en/2.9/extensions/plugin-interfaces.html#ckan.plugins.interfaces.IAuthFunctions).  This is done at the top of the UNAIDSPlugin class with the line `p.implements(p.IAuthFunctions)`
3) Add the function get_auth_functions to the UNAIDSPlugin class. The function should return a dictionary `{'organization_update_access': <auth fucntion from step 1>}` (A bit like registring a helper or action).
4) Then here in the process_dataset_transfer function you should call ```h.check_access('organization_update_access', {'org_to_allow_transfer_to': dataset['org_to_allow_transfer_to']})```  (h being ckan.lib.helpers)

ckanext/unaids/blueprints/unaids_dataset_transfer.py
```diff
@@ -0,0 +1,54 @@
...
+        'owner_org': dataset['org_to_allow_transfer_to'],
+        'org_to_allow_transfer_to': ''
+    })
+    toolkit.get_action('package_update')({
+        'model': model,
+        'session': model.Session,
+        'ignore_auth': True
+    }, dataset)
+
+    # return success
```
jonathansberry, line 39: Do these comments add much?  I'd probably either remove them, or make sure they say something about "why" rather than "what".

ckanext/unaids/blueprints/unaids_dataset_transfer.py
```diff
@@ -0,0 +1,54 @@
...
+    dataset = toolkit.get_action('package_show')({}, {'id': dataset_id})
+    
+    # validate user can transfer dataset
+    valid = helpers.check_organization_update_access(
+        dataset['org_to_allow_transfer_to']
+    )
+    if not valid:
+        return base.abort(403, _(u'You cannot carry out this action'))
+
+    # update dataset
```
jonathansberry, line 28: Do these comments add much? I'd probably either remove them, or make sure they say something about "why" rather than "what".

ckanext/unaids/helpers.py
```diff
@@ -82,3 +83,15 @@ def get_user_obj(field=""):
     Returns an attribute of the user object, or returns the whole user object.
     """
     return getattr(g.userobj, field, g.userobj)
+
+
+def check_organization_update_access(organization_id):
```
jonathansberry, line 88: Registering an tuh function similar to registering a helper once you have implement IAuthFunctions.  We should probably register properly as an auth function as mentioned above.

ckanext/unaids/tests/test_dataset_transfer.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+        # user_1 and user_3 should not be able to accept
+        # the transfer as they are not members of org_2
+        for user in [user_1, user_3]:
+            response = app.get(
+                url=transfer_dataset_url,
+                extra_environ={'REMOTE_USER': user['name']},
+                status=403 # expect forbidden error
+            )
```
jonathansberry, line 50: So are actually asserting here that a 403 is raised?  Could be good to do this using more standardised pytest syntax: 
```suggestion
        from werkzeug.exceptions import Forbidden
        with pytest.raises(Forbidden):
            response = app.get( 
                url=transfer_dataset_url,
￼                extra_environ={'REMOTE_USER': user['name']}
            )
```

jonathansberry: Import the exception at the top of the file obviously.

ckanext/unaids/tests/test_dataset_transfer.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+        # confirm dataset was moved to org_2
+        result = helpers.call_action(
+            'package_show',
+            id=dataset['id'],
+            context={'user': user_2['name']},
+        )
+        assert result['owner_org'] == org_2['id']
+        assert not result['org_to_allow_transfer_to']
```
jonathansberry, line 66: This is a great integration test.  I'd suggest writing a quick unit test for your validator and your auth function as well.  These should be dead simple in comparison because you just call the function directly with prepared dictionary values. There is no need to use the DB and try to call complicated flask end points with contexts etc...

ckanext/unaids/theme/templates/package/read.html
```diff
@@ -3,6 +3,25 @@
 {% set pkg = c.pkg_dict %}
 
 {% block package_description %}
+
+  {% if pkg.org_to_allow_transfer_to %}
+    <div class="alert alert-info" style="margin-top: -10px;">
+      {% if h.check_organization_update_access(pkg.org_to_allow_transfer_to) %}
```
jonathansberry, line 9: Your new auth function can be called properly with `h.check_access('organisation_editor`, {'org_id': pkg.org_to_allow_transfer_to})` I think.

ckanext/unaids/theme/templates/package/read.html
```diff
@@ -3,6 +3,25 @@
...
+        <p>{{ pkg.organization.title + _(' would like to transfer this dataset to your organisation') }}</p>
+        <br>
+        <a class="btn btn-default" href="{{ url_for('unaids_dataset_transfer.process_dataset_transfer', dataset_id=pkg.id) }}">
+          <i class="fa fa-refresh"></i>
+          <span> {{ _('Transfer dataset to ') +  h.get_organization(pkg.org_to_allow_transfer_to).display_name }}</span>
+        </a>
+      {% elif h.check_organization_update_access(pkg.owner_org) %}
```
jonathansberry, line 16: Your new auth function can be called properly with `h.check_access('organisation_editor`, {'org_id': pkg.owner_org})` I think.

jonathansberry (changes requested):
Fantastic work manoj. Really nearly there.  There are three main things I'm proposing:
- Register your own proper auth function for checking organisation access. Rather than doing it in a template helper. 
- A simple unit test for your auth function.  No need to be using factories or apps or dbs here.  Just call it directly with some good and bad hard coded dictionaries and check behaviour proceeds as expected. This way we'll clearly catch if something breaks expected behaviour in the future. 
- Likewise a simple unit test for your validator.

ckanext/unaids/blueprints/unaids_dataset_transfer.py
Manoj-nathwani: Great feedback. This is now done ✅

ckanext/unaids/tests/test_dataset_transfer.py
Manoj-nathwani: I copied this style of writing the test off another test elsewhere.
When expecting a 200 status I get an `AssertionError` rather than `Forbidden` which is equally as informative

```
for user in [user_1, user_3]:
    with pytest.raises(AssertionError):
        app.get(
            url=transfer_dataset_url,
            extra_environ={"REMOTE_USER": user["name"]},
            status=200,  # should return a 403
        )
```

It feels like sticking with my original `status=403 # expect forbidden error` code is more descriptive.
What do you think?

ckanext/unaids/auth.py
```diff
@@ -0,0 +1,21 @@
+# coding: utf
+import ckan.logic as logic
+
+
+def organization_update_access(context, data_dict=None):
```
jonathansberry, line 5: :ok_hand: :+1:

ckanext/unaids/theme/templates/package/read.html
```diff
@@ -6,14 +6,14 @@
 
   {% if pkg.org_to_allow_transfer_to %}
     <div class="alert alert-info" style="margin-top: -10px;">
-      {% if h.check_organization_update_access(pkg.org_to_allow_transfer_to) %}
+      {% if h.check_access('organization_update', {'id': pkg.org_to_allow_transfer_to}) %}
```
jonathansberry, line 9:
```suggestion
      {% if h.check_access('unaids_organization_update', {'id': pkg.org_to_allow_transfer_to}) %}
```

ckanext/unaids/theme/templates/package/read.html
```diff
@@ -6,14 +6,14 @@
...
         <p>{{ pkg.organization.title + _(' would like to transfer this dataset to your organisation') }}</p>
         <br>
         <a class="btn btn-default" href="{{ url_for('unaids_dataset_transfer.process_dataset_transfer', dataset_id=pkg.id) }}">
           <i class="fa fa-refresh"></i>
           <span> {{ _('Transfer dataset to ') +  h.get_organization(pkg.org_to_allow_transfer_to).display_name }}</span>
         </a>
-      {% elif h.check_organization_update_access(pkg.owner_org) %}
+      {% elif h.check_access('organization_update', {'id': pkg.owner_org}) %}
```
jonathansberry, line 16: As above...
```suggestion
      {% elif h.check_access('unaids_organization_update', {'id': pkg.owner_org}) %}
```

jonathansberry (changes requested):
Yeah great.  I'm just worried about there being an "organization_update" function specified somewhere else.  We don't want to unknowingly override it's behaviour in parts of the system. Suggestion is to rename to "unaids_organization_update".  Also, with a proper check access function registered, we can simplify the blueprint code a bit.

ckanext/unaids/tests/test_dataset_transfer.py
jonathansberry: I'd probably do...

Manoj-nathwani:
> Yeah great. I'm just worried about there being an "organization_update" function specified somewhere else. We don't want to unknowingly override it's behaviour in parts of the system. Suggestion is to rename to "unaids_organization_update". Also, with a proper check access function registered, we can simplify the blueprint code a bit.

yes! I was going to mention this. It feels like i'm reimplementing something which already exists however doesn't function as described. So it may be worth checking other features to make sure it's not broken as I could never get the `check_access` of `organization_update` to work

jonathansberry:
@Manoj-nathwani I've moved things around and added two stub tests, one for the validator and one for the authorize function.  They simply pass at the moment, but you'll want to devise some reasonable way of testing that they don't break for bad input, and pass for a few types of good input...

Manoj-nathwani:
```
============================= test session starts ==============================
platform linux2 -- Python 2.7.13, pytest-4.6.5, py-1.9.0, pluggy-0.13.1
rootdir: /usr/lib/adx/ckanext-unaids
plugins: ckan-0.0.12, pyfakefs-3.2, cov-2.7.1, freezegun-0.4.1, rerunfailures-8.0, split-tests-1.0.9
collected 7 items

../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_actions.py ..   [ 28%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_auth.py ..      [ 57%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_dataset_transfer.py . [ 71%]
                                                                         [ 71%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_plugin.py .     [ 85%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_validators.py . [100%]

==================== 7 passed, 169 warnings in 9.25 seconds ====================
```
Tests pass too ✅

jonathansberry (approved):

PR #87, opened 2020-11-16
Adx 506 webassets config

tomeksabala:
I have bundled all of unaids specific js files into a single webassets entry.
Also I've configured a dependency on core CKAN js so our js code is loaded in the right order (that's the primary reason why download all wasn't working before).

I have tried to do similar things to unaids css but I have failed to complete it. I don't know how to configure images and fonts to work with webassets config.


jonathansberry (approved):
Seems a little inefficient to load all the unaids javascript for all pages?  But it's going to be a negligible matter I imagine so happy to approve.

PR #94, opened 2020-12-08
Adx 464 make footer sticky

Manoj-nathwani:
# Problem
- Website footer floats in the middle of the page when there's little content to display

# Solution
- Set up a sticky footer
- See https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/

## before
image

## after
image

### DIFF
--- a/ckanext/unaids/theme/public/custom.css
+++ b/ckanext/unaids/theme/public/custom.css
@@ -209,7 +209,7 @@ section.resources H3{
 /* make footer sticky */
 .main{
   display: flex;
-  min-height: 75vh;
+  min-height: 60vh;
   flex-direction: column;
 }
 
--- a/ckanext/unaids/theme/public/custom.css
+++ b/ckanext/unaids/theme/public/custom.css
@@ -205,3 +205,14 @@ section.resources H3{
 .resource-title{
   margin-bottom: 10px;
 }
+
+/* make footer sticky */
+.main {
+  display: flex;
+  min-height: 75vh;
+  flex-direction: column;
+}
+
+footer {
+  flex: 1;
+}
--- a/ckanext/unaids/theme/public/custom.css
+++ b/ckanext/unaids/theme/public/custom.css
@@ -207,12 +207,12 @@ section.resources H3{
 }
 
 /* make footer sticky */
-.main {
+.main{
   display: flex;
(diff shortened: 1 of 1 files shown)


ckanext/unaids/theme/public/custom.css
```diff
@@ -205,3 +205,14 @@ section.resources H3{
 .resource-title{
   margin-bottom: 10px;
 }
+
+/* make footer sticky */
+.main{
+  display: flex;
+  min-height: 60vh;
```
jonathansberry, line 212: So this is saying make the website content (without footer) always at least 60% the height of the screen, then add the footer below that.  It therefore relies on the footer being more than 40% the height of the screen.  If someone has a particularly tall screen, the footer may still not stick to the bottom of the screen.  

This is definitely a step forwards, and I'd therefore just question the 60vh - why not 80vh?  or 90vh?  this would increase our chances of the footer always appearing at the bottom of the page for all screen sizes, at the cost of making some users scroll to see the whole footer.

What do you think? Can you give the rationale behind 60vh? Do you think it makes sense to bump the number up a bit?

jonathansberry (commented):
Just wondering about why 60vh - what would be the relative merits of pushing this up to say 80vh or  90vh?

ckanext/unaids/theme/public/custom.css
Manoj-nathwani: Yeah you're on the right track. There's no perfect way to make a sticky footer so you kinda need to gage a rough height you'd like to wait before placing it. I chose `60vh` after testing it with my 24" screen and it looked about right. I can't imagine anyone having anything larger or taller. The main focus is getting it looking good on everything under that size, especially mobile devices. So  yes - `60vh` was just something I arbitrarily chose after playing around with various numbers. Originally it was `100vh` which felt far too large. It's a function of how much fluff (headers, body padding, etc) we have on an empty page and what the max screen size we want to support before the footer becomes sticky. So i went with this.

Manoj-nathwani: tldr; Under 60 felt annoying on smaller devices. Higher than 60 felt like there was too much white space in our `.main` body which felt broken.

jonathansberry (approved):

PR #99, opened 2021-01-19
Adx 571 fix dataset-transfer ascii errors

Manoj-nathwani:
# Problem
Attempting to transfer a dataset with ascii characters causes an error
```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf4' in position 1: ordinal not in range(128)
  File "flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "py", line 734, in post
    pkg_dict = get_action(u'package_update')(context, data_dict)
  File "ckan/logic/__init__.py", line 473, in wrapped
    result = _action(context, data_dict, **kw)
  File "ckan/logic/action/update.py", line 330, in package_update
    item.after_update(context, data)
  File "usr/lib/ckan/extensions/ckanext-unaids/ckanext/unaids/plugin.py", line 134, in after_update
    recipient_org_id=org_to_allow_transfer_to[0]
  File "usr/lib/ckan/extensions/ckanext-unaids/ckanext/unaids/dataset_transfer/logic.py", line 74, in send_dataset_transfer_emails
    subject = '{} {}'.format(dataset['title'], _('Dataset Transfer'))
```

# Solution 
- Stop doing a `string.format()` on unicode strings
- Swap it out for a python unicode string concatenation

### DIFF
--- a/ckanext/unaids/dataset_transfer/logic.py
+++ b/ckanext/unaids/dataset_transfer/logic.py
@@ -2,7 +2,7 @@ import logging
 
 from ckan import model
 import ckan.plugins.toolkit as toolkit
-from ckan.lib import mailer
+import ckan.lib.mailer as mailer
 from ckan.lib.base import render_jinja2
 from ckan.common import config, _
 from ckanext.unaids.dataset_transfer.model import (
@@ -104,3 +104,4 @@ def send_dataset_transfer_emails(dataset_id, recipient_org_id):
                 dataset['id']
             )
         )
+    return emails_succeeded
--- a/ckanext/unaids/dataset_transfer/logic.py
+++ b/ckanext/unaids/dataset_transfer/logic.py
@@ -71,7 +71,7 @@ def send_dataset_transfer_emails(dataset_id, recipient_org_id):
     emails_succeeded = []
     for user in users_to_email:
         try:
-            subject = dataset['title'] + ' ' + _('Dataset Transfer')
+            subject = u'{} {}'.format(dataset['title'], _('Dataset Transfer'))
             body = render_jinja2(
                 'email/dataset_transfer.txt',
                 extra_vars={
--- a/ckanext/unaids/dataset_transfer/logic.py
+++ b/ckanext/unaids/dataset_transfer/logic.py
@@ -71,7 +71,7 @@ def send_dataset_transfer_emails(dataset_id, recipient_org_id):
     emails_succeeded = []
     for user in users_to_email:
         try:
-            subject = '{} {}'.format(dataset['title'], _('Dataset Transfer'))
+            subject = dataset['title'] + ' ' + _('Dataset Transfer')
             body = render_jinja2(
                 'email/dataset_transfer.txt',
                 extra_vars={
--- a/ckanext/unaids/tests/test_dataset_transfer.py
+++ b/ckanext/unaids/tests/test_dataset_transfer.py
@@ -97,7 +97,8 @@ class TestDatasetTransfer(object):
             )
 
     @mock.patch('ckan.lib.mailer.mail_user')
-    def test_send_dataset_transfer_emails(self, mocked_mail_user, app):
+    @mock.patch('ckan.lib.mailer._mail_recipient')
+    def test_send_dataset_transfer_emails(self, mocked_mail_recipient, mocked_mail_user, app):
         user_1, user_2 = [
             factories.User()
             for x in range(2)
@@ -112,10 +113,11 @@ class TestDatasetTransfer(object):
             title=u"Côte d'Ivoire Test Dataset",
             org_to_allow_transfer_to=org_2['id']
         )
-        send_dataset_transfer_emails(
+        emails_succeeded = send_dataset_transfer_emails(
             dataset_id=dataset['id'],
             recipient_org_id=org_2['id']
         )
+        assert emails_succeeded
         mocked_mail_user.assert_called()
         emailed_user_ids = [
             x[0][0].id  # recipient_user_id
@@ -137,12 +139,11 @@ class TestDatasetTransfer(object):
             type='test-schema',
             org_to_allow_transfer_to=org_2['id']
         )
-        expected_error = r'All DatasetTransferRequest emails failed *'
-        with pytest.raises(AssertionError, match=expected_error):
-            send_dataset_transfer_emails(
-                dataset_id=dataset['id'],
-                recipient_org_id=org_2['id']
-            )
+        emails_succeeded = send_dataset_transfer_emails(
+            dataset_id=dataset['id'],
+            recipient_org_id=org_2['id']
+        )
+        assert not emails_succeeded
 
     def test_dataset_transfer_request(self, app):
 
--- a/ckanext/unaids/tests/test_dataset_transfer.py
+++ b/ckanext/unaids/tests/test_dataset_transfer.py
@@ -97,8 +97,7 @@ class TestDatasetTransfer(object):
             )
 
     @mock.patch('ckan.lib.mailer.mail_user')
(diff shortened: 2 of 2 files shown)


Manoj-nathwani:
- I haven't managed to run the smtp server locally to confirm this works
- Testing via the UI no longer causes an error though

Manoj-nathwani:
Mades fixes to dataset_transfer tests too 🚀 
```
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_actions.py ..   [ 14%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_auth.py ..      [ 28%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_dataset_transfer.py . [ 35%]
.......                                                                  [ 85%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_plugin.py .     [ 92%]
../../../../adx/ckanext-unaids/ckanext/unaids/tests/test_validators.py . [100%]

=================== 14 passed, 334 warnings in 56.10 seconds ===================
```

Manoj-nathwani:
It's so hard to know what the email subject will be; unlike the body it's not base64

This is what the SMTP proxy says:
`b'Subject: =?utf-8?q?2021_Santa_Claus_Lists_C=C3=B4te_d=27Ivoire_Dataset_Transfer?='`

If I pass in the subject as `helooooooooooo` i get:
`b'Subject: =?utf-8?q?helooooooooooo?='`

and the new user registration email looks like:
`b'Subject: =?utf-8?q?New_Registration=3A_erhewjeet_=28wegewg=40ewgw=2Ecom=29?='`

Any ideas what this is encoded as to make it human readable?

This is quite useful:
https://www.webatic.com/html-entities-convertor

Manoj-nathwani:
It's so hard to know what the email subject will be; unlike the body it's not base64

This is what the SMTP proxy says:
`b'Subject: =?utf-8?q?2021_Santa_Claus_Lists_C=C3=B4te_d=27Ivoire_Dataset_Transfer?='`

If I pass in the subject as `helooooooooooo` i get:
`b'Subject: =?utf-8?q?helooooooooooo?='`

and the new user registration email looks like:
`b'Subject: =?utf-8?q?New_Registration=3A_erhewjeet_=28wegewg=40ewgw=2Ecom=29?='`

Any ideas what this is encoded as to make it human readable?

This is quite useful:
https://www.webatic.com/html-entities-convertor

Using the above html-entities-convertor I get:
`2021_Santa_Claus_Lists_Côte_d'Ivoire_Dataset_Transfer`

Which is likely correct as the account registration email decodes to:
`New_Registration:_erhewjeet_(wegewg@ewgw.com)`

The only confusing part is that spaces are converted to underscores, which is in the case of both my new code and the account registration code. So we can assume this to be correct ✅

ckanext/unaids/tests/test_dataset_transfer.py
```diff
@@ -136,12 +139,11 @@ def test_send_dataset_transfer_emails_errors(self, app):
...
-        with pytest.raises(AssertionError, match=expected_error):
-            send_dataset_transfer_emails(
-                dataset_id=dataset['id'],
-                recipient_org_id=org_2['id']
-            )
+        emails_succeeded = send_dataset_transfer_emails(
+            dataset_id=dataset['id'],
+            recipient_org_id=org_2['id']
+        )
+        assert not emails_succeeded
```
tomeksabala, line 146: You're not checking that not e-mails got send out in this case. In order to do that you'll need two different mocks.
I suggest using it with `with` keyword instead of annotation:
```
with mock.patch('ckan.lib.mailer.mail_user') as mail_user_mock:
     [test code]
```
Also this could be split into two tests:
1. Which tests if e-mail with proper text are send on dataset transfer request
2. Verify if e-mail are not sent twice.

ckanext/unaids/dataset_transfer/logic.py
```diff
@@ -71,7 +71,7 @@ def send_dataset_transfer_emails(dataset_id, recipient_org_id):
     emails_succeeded = []
     for user in users_to_email:
         try:
-            subject = '{} {}'.format(dataset['title'], _('Dataset Transfer'))
+            subject = dataset['title'] + ' ' + _('Dataset Transfer')
```
tomeksabala, line 74: I don't think it's a good idea on how to do it. In python2 unfortunately you need to explicitly for unicode space for strings:
```
 subject = u'{} {}'.format(dataset['title'], _('Dataset Transfer'))
 ```
 Should do just enough. The subject will be properly encoded later down the line with `ckan/ckan/lib/mailer.py`:
 ```
    subject = Header(subject.encode('utf-8'), 'utf-8')
 ````

**Review by tomeksabala — CHANGES_REQUESTED** [review]
Almost there. The reason this bug exists is due to poorly handled encoding issues.
It's not an easy thing to grasp, especially in python2 (in python3 the defaults are set a little better so it's harder to make mistakes).

I recommend reading this article: https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/ (I still tend to come back to it from time to time).

File: ckanext/unaids/tests/test_dataset_transfer.py [hunk 3]
```diff
@@ -97,31 +97,49 @@ def test_get_org_admins_with_no_users_in_org(self, app):
...
-            for x in range(2)
-        ]
-        org_1, org_2 = [
-            factories.Organization(user=user)
-            for user in [user_1, user_2]
-        ]
+    @mock.patch('ckan.lib.mailer._mail_recipient')
+    def test_send_dataset_transfer_emails(self, mocked_mail_recipient, mocked_mail_user, app):
+        user_1 = factories.User(email='user_1@example.com')
+        user_2 = factories.User(email='user_2@example.com')
```
tomeksabala, line 103: why this approach instaed of the older one:
```
        user_1, user_2 = [
            factories.User()
            for x in range(2)
        ]
```
You're asiging spcific e-mails to those test users but they're not used afterwards. If it's the case it'd be better to leave them with default values.

ckanext/unaids/tests/test_dataset_transfer.py
```diff
@@ -97,31 +97,49 @@ def test_get_org_admins_with_no_users_in_org(self, app):
...
-    def test_send_dataset_transfer_emails(self, mocked_mail_user, app):
-        user_1, user_2 = [
-            factories.User()
-            for x in range(2)
-        ]
-        org_1, org_2 = [
-            factories.Organization(user=user)
-            for user in [user_1, user_2]
-        ]
+    @mock.patch('ckan.lib.mailer._mail_recipient')
```
tomeksabala, line 100: With  a quick check it looks like the mail_receipient patch is not used.
You shouldn't be mocking protected (marked with `_` prefix) functions/fields in general.
After removing this mock test passed just fine. Please have a look.

tomeksabala (changes requested):

ckanext/unaids/tests/test_dataset_transfer.py
Manoj-nathwani: I couldn't get tests passing without doing this, otherwise I would have left it. I know it doesn't make too much sense but my guess was that you need to specify an email address as the factory wasn't doing a good enough job to avoid errors.

tomeksabala (approved):

PR #101, opened 2021-02-15
Create robot.txt

Manoj-nathwani:
Stop crawler bots spamming us 🕷️ 🔨 

As discussed: https://fjelltopp.slack.com/archives/CFD6YJURG/p1613369667002900


ckanext/unaids/theme/public/robot.txt
```diff
@@ -0,0 +1,2 @@
+Disallow: /*/organization/*
+Disallow: /*/dataset/*
```
jonathansberry, line 2: Ok, @fulior recommends disallowing all traffic and then allow traffic to certain areas of the ADR.  I think I would recommend the following for now, and then in time we may look at allowing datasets them selves to be indexed for the purpose of google searches.  This is not important at the moment though.
```suggestion
User-agent: *
Disallow: /*
Allow: /*/pages/
Crawl-Delay: 10
```
Doing this allows the home page and all ckanext-pages content to be crawled, but nothing else.

jonathansberry (changes requested):

jonathansberry (approved):

PR #109, opened 2021-04-13
removed duplicate extstorage helpers

Manoj-nathwani:
Remove duplicate helpers copy/pasted from `ckanext.external_storage.helpers` and directly call them from unaids


ckanext/unaids/plugin.py
```diff
@@ -102,10 +100,10 @@ def get_helpers(self):
...
             u'get_user_obj': get_user_obj,
             u'get_all_organizations': get_all_organizations,
-            u'extstorage_resource_authz_scope': get_extstorage_resource_authz_scope,
-            u'extstorage_resource_storage_prefix': get_extstorage_resource_storage_prefix,
-            u'extstorage_resource_filename': get_extstorage_resource_filename,
-            u'max_resource_size': get_max_resource_size
+            u'extstorage_resource_authz_scope': extstorage_helpers.resource_authz_scope,
+            u'extstorage_resource_storage_prefix': extstorage_helpers.resource_storage_prefix,
+            u'extstorage_resource_filename': extstorage_helpers.resource_filename,
+            u'max_resource_size': uploader.get_max_resource_size()
```
tomeksabala, line 106: I would keep the value as function not a value. This is very confusing. You'd expect to get it called like:
```
helpers.max_resource_size()
```

ckanext/unaids/theme/templates/scheming/form_snippets/upload.html
```diff
@@ -3,7 +3,7 @@
 <div
     id="FileInputComponent"
     data-lfsServer="{{ h.extstorage_server_url() }}"
-    data-maxResourceSize="{{ h.max_resource_size() }}"
+    data-maxResourceSize="{{ h.max_resource_size }}"
```
tomeksabala, line 6: See my other comment to `helpers.py`

tomeksabala (changes requested):
Nice! 55 lines of code removed net.

Please update the `max_resource_size` helper def. it should be a function not a value.

PR #111, opened 2021-04-15
get-authz-token-before-file-upload

Manoj-nathwani:
https://fjelltopp.atlassian.net/jira/software/projects/ADX/boards/4?selectedIssue=ADX-647

# Problem
- We were requesting the authz token on page load
- This meant if you open the file upload page, leave it open for 30mins, the token would expire when you finally got around to selecting a file to upload

# Solution
- We now request a new auth token at the time of actually uploading something

### DIFF
--- a/ckanext/unaids/react/components/FileInputComponent/src/FileUploader.js
+++ b/ckanext/unaids/react/components/FileInputComponent/src/FileUploader.js
@@ -14,42 +14,44 @@ export default function FileUploader({
             '/api/3/action/authz_authorize',
             { scopes: `obj:ckan/${datasetId}/*:write` },
             { withCredentials: true }
-        ).then(res => res.data.result.token)
-
-    const handleFileSelected = async inputFile => {
-        if (!inputFile) return;
-        setUploadProgress({ loaded: 0, total: 1 });
-        const file = data.open(inputFile);
-        getAuthToken()
-            .then(async authToken => {
-                const client = new Client(lfsServer, authToken, ['basic']);
-                await client.upload(file, orgId, datasetId, progress => {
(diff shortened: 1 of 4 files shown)


ckanext/unaids/react/components/FileInputComponent/src/FileUploader.js
```diff
@@ -1,37 +1,55 @@
...
     setUploadProgress, setUploadFileName, setHiddenInputs,
     setUploadError
 }) {
 
+    const getAuthToken = () =>
+        axios.post(
+            '/api/3/action/authz_authorize',
+            { scopes: `obj:ckan/${datasetId}/*:write` },
+            { withCredentials: true }
+        ).then(res => res.data.result.token)
```
tomeksabala, line 17: I think you're not handling `authz_authorize` error anywhere. Or is it going to be caught by:
``` 
            .catch(error => {
                setUploadError({
                    error: ckan.i18n._('Server Error'),
                    description: ckan.i18n._('An unknown server error has occurred.')
                });
                throw error;
            })
```
later down the line?

tomeksabala (changes requested):

ckanext/unaids/react/components/FileInputComponent/src/FileUploader.js
Manoj-nathwani: Yep that's the case

tomeksabala: @Manoj-nathwani Why not do it in the same way as in batch upload then?
image

tomeksabala (approved):
Ok!

PR #112, opened 2021-04-15
Adx 648 request authz token before each bulked file upload

Manoj-nathwani:
https://fjelltopp.atlassian.net/jira/software/projects/ADX/boards/4?selectedIssue=ADX-647

# Problem
- We were requesting the authz token before uploading any files, then recycling the same token for every file upload
- The issue is that if a user uploads 20 files, each taking 2mins to upload, then the last few files would fail to upload as the auth token would have expired

# Solution
- We now request a new auth token before uploading each file

### DIFF
--- a/ckanext/unaids/react/components/BulkFileUploaderComponent/index.test.js
+++ b/ckanext/unaids/react/components/BulkFileUploaderComponent/index.test.js
@@ -53,36 +53,38 @@ const selectFilesToUpload = async (elementTestId, files) => {
     fireEvent.drop(component);
   });
 }
-const uploadFilesAndCreateResources = async successfulFileUploads => {
+const uploadFilesAndCreateResources = async (validFileUploads, invalidFileUpload) => {
   fireEvent.click(screen.getByTestId('UploadFilesButton'));
   await screen.findByText('Uploads Complete');
   const filesUploaded = await screen.findAllByText('Uploaded');
-  expect(filesUploaded).toHaveLength(successfulFileUploads.length);
+  expect(filesUploaded).toHaveLength(validFileUploads.length);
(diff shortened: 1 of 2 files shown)


Manoj-nathwani:
Once merged, carry out this: https://fjelltopp.atlassian.net/browse/ADX-649?focusedCommentId=10763

ckanext/unaids/react/components/BulkFileUploaderComponent/index.test.js
```diff
@@ -53,35 +53,38 @@ const selectFilesToUpload = async (elementTestId, files) => {
...
-const uploadFilesAndCreateResources = async successfulFileUploads => {
+const uploadFilesAndCreateResources = async (validFileUploads, invalidFileUpload) => {
   fireEvent.click(screen.getByTestId('UploadFilesButton'));
   await screen.findByText('Uploads Complete');
   const filesUploaded = await screen.findAllByText('Uploaded');
-  expect(filesUploaded).toHaveLength(successfulFileUploads.length);
+  expect(filesUploaded).toHaveLength(validFileUploads.length);
+  const requestsForAuthToken = validFileUploads.length + invalidFileUpload.length;
+  const requestsForUploadFile = requestsForAuthToken - invalidFileUpload.length;
   expect(mockedAxiosPost).toHaveBeenCalledTimes(
```
tomeksabala, line 63: Please look at my comment to the other PR about the mock name for `mockedAxiosPost`

tomeksabala (changes requested):
This looks good. I think it'd be worth if you consider updating one mock name (it's not direct part of this PR but it's very close in scope so let's add it here).

tomeksabala (approved):
Looks good to me! :)

PR #114, opened 2021-04-16
fix-uploads-completed-timing

Manoj-nathwani:
# Before
Before all files have finished uploading, we wrongly state Uploads Complete

image

# After
Now we correctly show it's still uploading

image

Then once completed, it says Uploads Complete

image


tomeksabala (approved):
Looks good to me! :)

PR #115, opened 2021-04-16
fix-handleNetworkError-undefined

Manoj-nathwani:
This was a bad copy/paste from the bulk file uploader where `handleNetworkError` was never defined in this file


tomeksabala (approved):
Looks good to me! :)

PR #116, opened 2021-04-20
add delete resource button to dataset_show

Manoj-nathwani:
## Problem
The `Delete` button has gone missing on the dataset_show page

image

## Solution
image

Originally `ckanext-file_uploader_ui` was adding the `Delete` button to `resource_item_explore_links`

I've copied the `<li>` from here: https://github.com/fjelltopp/ckanext-file_uploader_ui/blob/development/ckanext/file_uploader_ui/templates/package/snippets/resource_item.html#L37-L44


jonathansberry (approved):
Looks good to me

PR #117, opened 2021-05-07
ADX-657 Added new Imperial logo. Fixed deprecated <center>

tomeksabala:
image


Manoj-nathwani (changes requested):
- You can set a bootstrap class rather than adding inline styles for most of these.
- There's also a bunch of other `<center>` tags in the document you may also want to change
-  A screenshot would also be nice

tomeksabala:
>     * You can set a bootstrap class rather than adding inline styles for most of these.
Valid point. I've updated it.
> 
>     * There's also a bunch of other `<center>` tags in the document you may also want to change
I have only changed nearby divs of changed row. It's just a logo update and the rest of divs need to wait.
> 
>     * A screenshot would also be nice
Attached.

Manoj-nathwani (approved):
👍

PR #118, opened 2021-05-07
ADX-656 core and extra resources

Manoj-nathwani:
See https://github.com/fjelltopp/ckanext-scheming/pull/55


tomeksabala:
@Manoj-nathwani Why do we need to change the test_schema of unaids extenstion?

Manoj-nathwani:
> @Manoj-nathwani Why do we need to change the test_schema of unaids extenstion?

The tests in the file I was editing already were referencing this schema file.
I needed to add an expected resource constraint so my changes would have a concept of a required file.

PR #126, opened 2021-05-17
Adx 658 email user search

jonathansberry:
This change accompanies the PR for ADX-658 in ckanext-emailasusername. 

It makes some small changes to the UI so that the placeholder of user search inputs indicates that you can search by email. 

For instance the restricted fields of a resource...

image

image

Have I missed any other places in the UI where people commonly search for users?

### DIFF
--- a/ckanext/unaids/i18n/extra_translations.py
+++ b/ckanext/unaids/i18n/extra_translations.py
@@ -48,5 +48,5 @@ extra_translations = [
     _('A spectrum file used to run the Naomi model'),
     _('License definitions and additional information can be found at http://opendefinition.org/'),
     _('Reorder resources'),
-
+    _('Organisation Name / ID')
 ]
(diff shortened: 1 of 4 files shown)


ckanext/unaids/i18n/extra_translations.py
```diff
@@ -48,5 +48,5 @@
     _('A spectrum file used to run the Naomi model'),
     _('License definitions and additional information can be found at http://opendefinition.org/'),
     _('Reorder resources'),
-
+    _('Organisation Name / ID')
```
jonathansberry, line 51: The string "Username / Email" doesn't need to be added as a string here because it is referenced directly from the code in other templates and will be picked up from there.  Of course I could add it in here for completeness if this is deemed important.

tomeksabala (approved):

PR #129, opened 2021-05-19
smoother react component loading flow

Manoj-nathwani:
# Problem
- The react single file uploader renders content before all states have been updated
- This causes the url field to display regardless if there's already a url/file set which causes an odd lag effect
- You can see it here: https://youtu.be/3mAPjIpSd1g

# Solution
- We're now reading the raw html sitting inside the component pre-render and storing it. We then display it until the UseEffect has completed running and all states have been updated.

### DIFF
--- a/ckanext/unaids/react/components/FileInputComponent/index.js
+++ b/ckanext/unaids/react/components/FileInputComponent/index.js
@@ -9,6 +9,7 @@ const getAttr = key => {
   return ['None', ''].includes(val) ? null : val;
 };
 const
+  loadingHtml = componentElement.innerHTML,
   maxResourceSize = parseInt(getAttr('maxResourceSize')),
   lfsServer = getAttr('lfsServer'),
   orgId = getAttr('orgId'),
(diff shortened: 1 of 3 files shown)


Manoj-nathwani:
# Update
- I did some research in the most ideal way to make this flow work
- The main blocker here was that we don't have access to `ckan.i18n` through react early enough to display the loading containers. So i'm having the jinja2 template render it and then we're padding it as raw html to react which displays it until it feels ready to show the fully rendered content.

I tried updating the webassets.yaml:
```
FileInputComponentScripts:
    contents:
        - frictionless-js.js
        - build/FileInputComponent.js
    output: unaids/%(version)s_FileInputComponent.js
    extra:
        preload:
            - base/main
```

However it didn't help as it just loads our component equally as slowly: https://youtu.be/qzI23ds1AOI

# Conclusion
- This PR is still the best way to achieve a clean, non-laggy UI

tomeksabala (approved):

PR #131, opened 2021-05-20
ADX-595 Upgrade authz-service and blob-storage versions.

tomeksabala:
This PR includes two changes:
1. It switches to use of ckan action `authorize` which is possible with latest fix to ckanext-authz-service https://github.com/datopian/ckanext-authz-service/issues/24
2. It uses newer version of ckanext-blobstoragev0.8.0 (+fjelltopp custom changes to support auth-service form 1)) which required some renaming due to external_storage -> blob_storage changes.


Manoj-nathwani (approved):
Looks good 👍

PR #140, opened 2021-06-29
ADX-715 dataset releases frontend updates

Manoj-nathwani:
[Jira ticket ADX-715](https://fjelltopp.atlassian.net/jira/software/projects/ADX/boards/4?selectedIssue=ADX-715)

## Releases list view is now responsive and mobile friendly
▶️ [View video of responsiveness](https://youtu.be/CzaulymHjx0)

| before     | after    |
|-------|-----------------------|
| image | image
| image | image |

## underlines have now been changed/removed
I tried replacing the `<u>` with a `<b>` however we don't seem to support bold with the font we're using anywhere on the ADR, which is a little odd. The UI isn't hugely changed by removing underlines anyway, so we can keep it like this until we look into the bold issue.

| before     | after    |
|-------|-----------------------|
| image | image
| image | image

### DIFF
--- a/ckanext/unaids/theme/templates/package/releases/list.html
+++ b/ckanext/unaids/theme/templates/package/releases/list.html
@@ -17,10 +17,10 @@
         </p>
     {% else %}
         <div class="btn-group">
-            <a href="{{ urls.create_from_latest_version }}" class="btn btn-danger">
+            <a href="{{ urls.create_from_latest_version }}" class="btn btn-primary">
                 <span>{{ _('Add Release') }}</span>
             </a>
-            <button type="button" class="btn btn-danger dropdown-toggle" data-toggle="dropdown">
+            <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown">
                 <span class="caret"></span>
             </button>
             <ul class="dropdown-menu">
--- a/ckanext/unaids/theme/templates/package/releases/list.html
+++ b/ckanext/unaids/theme/templates/package/releases/list.html
@@ -26,16 +26,12 @@
             <ul class="dropdown-menu">
                 <li>
                     <a href="{{ urls.create_from_latest_version }}">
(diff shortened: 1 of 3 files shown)


ckanext/unaids/theme/templates/package/releases/list.html
```diff
@@ -17,25 +17,21 @@
...
-                        <span>{{ _('of the dataset') }}</span>
+                        <span>{{ _('From the latest version of the dataset') }}</span>
                     </a>
                 </li>
                 <li>
                     <a href="{{ urls.create_from_an_older_version }}">
-                        <span>{{ _('From a') }} </span>
-                        <u>{{ _('past version') }} </u>
-                        <span>{{ _('of the dataset') }}</span>
+                        <span>{{ _('From a past version of the dataset') }}</span>
```
tomeksabala, line 34: FYI This fixed another translation issue too:
Older version would make translator to deal with 3 separate strings:
1. From a
2. past version
3. of the dataset
This would be really hard to deal with. If you use _('From a <u>past version</u> of a dataset') you present the translator with a  bit of html (which is totally fine) but you also give a full context to them.

ckanext/unaids/theme/templates/package/releases/list.html
```diff
@@ -17,25 +17,21 @@
         </p>
     {% else %}
         <div class="btn-group">
-            <a href="{{ urls.create_from_latest_version }}" class="btn btn-danger">
-                <span>{{ _('Add Release') }}</span>
+            <a href="{{ urls.create_from_latest_version }}" class="btn btn-primary">
+                <i class="fa fa-plus-square"></i> <span>{{ _('Add Release') }}</span>
```
tomeksabala, line 21: I have added the plus sign here myself as we use it for all other "Add" buttons.

tomeksabala (approved):

tomeksabala (approved):

PR #145, opened 2021-07-02
ADX-717 fix last updated on resource uploads

Manoj-nathwani:
# Problem
- When uploading urls/files through the single or bulk file uploaders, we are never setting the `last_updated` field on the resource

# Solution
- Using `before_create` and `before_update` set `last_updated` to a timestamp if (and only if) a change has been made to the file or url fields which involves tracking `['url_type', 'lfs_prefix', 'sha256', 'size', 'url']` for any changes.
- The react single file uploader has also been updated to us the `existingResourceData.url` in the resource `url` field, rather than `fileName`.

### DIFF
--- a/ckanext/unaids/plugin.py
+++ b/ckanext/unaids/plugin.py
@@ -256,7 +256,8 @@ def _giftless_upload(context, resource, current=None):
 
 
 def _update_resource_last_modified_date(resource, current=None):
-    current = current or {}
+    if current is None:
+        current = {}
     for key in ['url_type', 'lfs_prefix', 'sha256', 'size', 'url']:
         current_value = str(current.get(key) or '')
         resource_value = str(resource.get(key) or '')
--- a/ckanext/unaids/plugin.py
+++ b/ckanext/unaids/plugin.py
@@ -174,7 +174,7 @@ class UNAIDSPlugin(p.SingletonPlugin, DefaultTranslation):
     def before_update(self, context, current, resource):
         if _data_dict_is_resource(resource):
             _giftless_upload(context, resource, current=current)
-            _update_resource_last_modified_date(resource, current)
+            _update_resource_last_modified_date(resource, current=current)
         return resource
 
 
@@ -255,7 +255,8 @@ def _giftless_upload(context, resource, current=None):
             })
 
 
-def _update_resource_last_modified_date(resource, current={}):
+def _update_resource_last_modified_date(resource, current=None):
+    current = current or {}
     for key in ['url_type', 'lfs_prefix', 'sha256', 'size', 'url']:
         current_value = str(current.get(key) or '')
         resource_value = str(resource.get(key) or '')
--- a/ckanext/unaids/plugin.py
+++ b/ckanext/unaids/plugin.py
@@ -168,11 +168,13 @@ class UNAIDSPlugin(p.SingletonPlugin, DefaultTranslation):
     def before_create(self, context, resource):
         if _data_dict_is_resource(resource):
             _giftless_upload(context, resource)
+            _update_resource_last_modified_date(resource)
         return resource
 
     def before_update(self, context, current, resource):
         if _data_dict_is_resource(resource):
             _giftless_upload(context, resource, current=current)
+            _update_resource_last_modified_date(resource, current)
         return resource
 
 
@@ -251,6 +253,15 @@ def _giftless_upload(context, resource, current=None):
                 'url': attached_file.filename,
                 'lfs_prefix': lfs_prefix
             })
+
+
+def _update_resource_last_modified_date(resource, current={}):
+    for key in ['url_type', 'lfs_prefix', 'sha256', 'size', 'url']:
+        current_value = str(current.get(key) or '')
+        resource_value = str(resource.get(key) or '')
+        if current_value != resource_value:
+            resource['last_modified'] = datetime.datetime.utcnow()
+            return
 
 
 def _get_upload_authz_token(context, dataset_name, org_name):
--- a/ckanext/unaids/tests/test_plugin.py
+++ b/ckanext/unaids/tests/test_plugin.py
@@ -9,50 +9,59 @@ from ckanext.unaids.plugin import (
 )
 
 
+def _resource_dict(resource_type='file'):
+    if resource_type == 'file':
+        return {
+            'url_type': 'upload',
+            'url': 'file.csv',
+            'lfs_prefix': 'prefix',
+            'sha256': 'sha256',
+            'size': 100,
+        }
+    elif resource_type == 'link':
+        return {
+            'url_type': '',
+            'url': 'http://example.com'
+        }
+    else:
+        raise ValueError("Unsupported resource type %s", resource_type)
+
+
 @pytest.fixture
 def resource_with_file():
-    return {
-        'url_type': 'upload',
-        'url': 'file.csv',
-        'lfs_prefix': 'prefix',
-        'sha256': 'sha256',
-        'size': 100,
-    }
+    return _resource_dict(resource_type='file')
 
 
 @pytest.fixture
 def resource_with_link():
-    return {
-        'url_type': '',
-        'url': 'http://example.com'
-    }
+    return _resource_dict(resource_type='link')
 
 
 @pytest.fixture
 def resource_with_updated_file():
-    updated_resource = resource_with_file()
-    updated_resource.update({'url': 'file2.csv'})
+    updated_resource = _resource_dict(resource_type='file')
+    updated_resource['url'] = 'file2.csv'
     return updated_resource
 
 
 @pytest.fixture
 def resource_with_updated_link():
-    updated_resource = resource_with_link()
-    updated_resource.update({'url': 'http://example2.com'})
+    updated_resource = _resource_dict(resource_type='link')
+    updated_resource['url'] = 'http://example2.com'
     return updated_resource
 
 
 @pytest.fixture
 def resource_with_file_and_updated_metadata():
-    updated_resource = resource_with_file()
-    updated_resource.update({'key': 'value'})
+    updated_resource = _resource_dict(resource_type='file')
+    updated_resource['description'] = 'updated-description'
     return updated_resource
 
 
 @pytest.fixture
 def resource_with_link_and_updated_metadata():
-    updated_resource = resource_with_link()
-    updated_resource.update({'key': 'value'})
+    updated_resource = _resource_dict(resource_type='link')
+    updated_resource['description'] = 'updated-description'
     return updated_resource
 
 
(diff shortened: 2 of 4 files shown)


ckanext/unaids/plugin.py
```diff
@@ -168,11 +168,13 @@ def after_update(self, context, pkg_dict):
     def before_create(self, context, resource):
         if _data_dict_is_resource(resource):
             _giftless_upload(context, resource)
+            _update_resource_last_modified_date(resource)
         return resource
 
     def before_update(self, context, current, resource):
         if _data_dict_is_resource(resource):
             _giftless_upload(context, resource, current=current)
+            _update_resource_last_modified_date(resource, current)
```
tomeksabala, line 177: Better call kwarg explicitly:
```
_update_resource_last_modified_date(resource, current=current)
```

ckanext/unaids/plugin.py
```diff
@@ -253,6 +255,15 @@ def _giftless_upload(context, resource, current=None):
             })
 
 
+def _update_resource_last_modified_date(resource, current={}):
```
tomeksabala, line 258: Default args values should be immutable. https://docs.python-guide.org/writing/gotchas/
Better:
```
def _update_resource_last_modified_date(resource, current=None):
  if not current:
    current = {}
[...]
```

ckanext/unaids/plugin.py
```diff
@@ -253,6 +255,15 @@ def _giftless_upload(context, resource, current=None):
...
+    for key in ['url_type', 'lfs_prefix', 'sha256', 'size', 'url']:
+        current_value = str(current.get(key) or '')
```
tomeksabala, line 260: `.get()` can return custom default if key not present. 
Better:
`str(current.get(key, ''))`

ckanext/unaids/tests/test_plugin.py
```diff
@@ -30,3 +33,61 @@ def test_geojson_format_guessed_correctly(self):
...
+    '''Tests for the ckanext.plugin._update_resource_last_modified_date module.
+
+    Make sure we are setting the last_modified to the resource dict
+    whenever a resource is created or edited.
+    '''
+
+    URL_TYPE_FILE_UPLOAD = 'upload'
+    URL_TYPE_LINK = ''
+
+    def test_resource(self, url_type):
```
tomeksabala, line 48: this method breaks the tests. Prefix `test_` in a method name makes pytest to pick this up as a test and adds it to the suite.
https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery

ckanext/unaids/tests/test_plugin.py
```diff
@@ -30,3 +33,61 @@ def test_geojson_format_guessed_correctly(self):
...
+                'url': 'file.csv',
+                'lfs_prefix': 'prefix123',
+                'sha256': 'sha256',
+                'size': 100,
+            })
+        elif url_type == self.URL_TYPE_LINK:
+            resource_dict.update({
+                'url': 'http://example.com'
+            })
+        return resource_dict
```
tomeksabala, line 61: This should be two fixtures: 
https://docs.pytest.org/en/6.2.x/fixture.html#what-fixtures-are
1.
```
@pytest.fixture
def resource_with_file():
[...]
```
```
@pytest.fixture
def resource_with_link():
[...]
```


Then:
1.
```
    def test_null_to_file(self, resource_with_file):
        _update_resource_last_modified_date(resource=resource_with_file)
        assert 'last_modified' in resource
```
2.
```
    def test_file_to_url(self, resource_with_file, resource_with_link):
        current = self.test_resource(self.URL_TYPE_FILE_UPLOAD)
        resource = self.test_resource(self.URL_TYPE_LINK)
        _update_resource_last_modified_date(resource_with_link, current=resource_with_file)
        assert 'last_modified' in resource
```
etc.

tomeksabala (changes requested):
1. (!!!) This PR is breaking the tests (more in inline comments)
2. Small changes to the code updates suggested
3. Bigger refactor to the test code needed.

ckanext/unaids/plugin.py
tomeksabala: After discussion on meets we've decided that it's wrong suggestion because:
```
>>> str(None)
'None'
>>> str({'key': None}.get('key', ''))
'None'
>>> str({'key': None}.get('key') or '')
''
```

ckanext/unaids/plugin.py
```diff
@@ -255,7 +255,8 @@ def _giftless_upload(context, resource, current=None):
             })
 
 
-def _update_resource_last_modified_date(resource, current={}):
+def _update_resource_last_modified_date(resource, current=None):
+    current = current or {}
```
tomeksabala, line 259: That's an anti pattern. Current might evaluate to false in many other cases then just `None` leading to hard to debug errors.
I suggest using the standard:
```
if current is None:
  current = {}
```

One example discussion can be found here in the comments: https://stackoverflow.com/a/13710667/3432120

ckanext/unaids/tests/test_plugin.py
```diff
@@ -42,52 +42,85 @@ class TestResourceLastModified(object):
...
-        _update_resource_last_modified_date({}, resource)
-        assert 'last_modified' not in resource
-
-    def test_url_to_file(self):
-        current = self.test_resource(self.URL_TYPE_LINK)
-        resource = self.test_resource(self.URL_TYPE_FILE_UPLOAD)
-        _update_resource_last_modified_date(resource, current)
-        assert 'last_modified' in resource
+    @pytest.fixture
+    def resource_with_file(autouse=True):
```
tomeksabala, line 46: If you'd like to enable autouse it should go into decorator config:
```
@pytest.fixture(autouse=True)
def res_with_file():
[...]
```
I don't think it's a good idea to use autouse here anyway as you're not **always** using this fixtures in tests.
My suggestion: remove autouse kwarg.

ckanext/unaids/tests/test_plugin.py
```diff
@@ -42,52 +42,85 @@ class TestResourceLastModified(object):
...
+    def test_file_to_null(self, resource_with_file):
+        null_resource = {}
+        _update_resource_last_modified_date(
+            null_resource, current=resource_with_file
+        )
+        assert 'last_modified' not in resource_with_file
+
+    def test_file_to_file(self, resource_with_file):
+        updated_resource = resource_with_file.copy()
+        updated_resource.update({'url': 'file2.csv'})
```
tomeksabala, line 75: Simply create another fixture
```
@fixture
def updated_resource_with_file():
 # returning dict with new 'url' value
```

ckanext/unaids/tests/test_plugin.py
```diff
@@ -42,52 +42,85 @@ class TestResourceLastModified(object):
...
+        )
+        assert 'last_modified' in updated_resource
+
+    def test_file_to_link(self, resource_with_file, resource_with_link):
+        _update_resource_last_modified_date(
+            resource_with_link, current=resource_with_file
+        )
+        assert 'last_modified' in resource_with_link
+
+    def test_changing_random_metadata_in_file(self, resource_with_file):
```
tomeksabala, line 87: There's nothing random in this test. Name `test_changing_metadata_in_file` is shorter and exactly explains the arrange stage of the test.

ckanext/unaids/tests/test_plugin.py
```diff
@@ -42,52 +42,85 @@ class TestResourceLastModified(object):
...
+        }
+
+    @pytest.fixture
+    def resource_with_link(autouse=True):
+        return {
+            'url_type': '',
+            'url': 'http://example.com'
+        }
+
+    def test_null_to_file(self, resource_with_file):
```
tomeksabala, line 62: Could you update all test method to include the desired outcome of the test?
This will document the test and make test reports (and failure reports!) more descriptive. The latter is especially important if you get an error report from CI (github actions, jenkins etc.)

Generally: `test_[explain init situation]_should_[provide expected result]`
here: `test_null_to_file_should_update_last_modified_datetime`

This should be applied to all test cases in the class.

tomeksabala (changes requested):

ckanext/unaids/tests/test_plugin.py
Manoj-nathwani: I didn't intend on using it originally, however without it I get the below error:
```
_____________ ERROR collecting ckanext/unaids/tests/test_plugin.py _____________
Could not determine arguments of <bound method ?.resource_with_file of <ckanext.unaids.tests.test_plugin.TestResourceLastModified object at 0x7fcd0c5da7d0>>: invalid method signature
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
```

ckanext/unaids/tests/test_plugin.py
Manoj-nathwani: I'd rather not as the fixtures would only be used once so there's no need to take it out into it's own function. I also want to be super clear that i'm taking the original dictionary and only updating a single field, rather than the entire thing.

ckanext/unaids/tests/test_plugin.py
tomeksabala: You're defining the method inside a class, so first argument needs to be `self`. You're just doing a really weird thing here which might be described like this:
```
class Foo():
   def bar(self=True):  <<<<<< def my_fixture(autouse=True)
         pass
```
There's no need to define the fixture in the scope of the class. Just move it up to the module level and you're good.

ckanext/unaids/tests/test_plugin.py
tomeksabala: Making it another fixture will look just so much better. And it'll be easier to reuse the test input in the future.
If you really want to keep the intent of overwriting just a single field you can use:
```
@fixture
def original():
    return {'url': 'foo', 'description': 'bar'}

@fixture
def updated(original):
     return original.copy().update({'description': 'baz'})
```

ckanext/unaids/tests/test_plugin.py
```diff
@@ -30,3 +33,94 @@ def test_geojson_format_guessed_correctly(self):
...
+        _update_resource_last_modified_date(
+            resource_with_file, current=resource_with_link
+        )
+        assert 'last_modified' in resource_with_file
+
+    def test_changing_metadata_in_link_should_not_update_last_modified_datetime(self, resource_with_link):
+        updated_resource = resource_with_link.copy()
+        updated_resource.update({'key': 'value'})
+        _update_resource_last_modified_date(
+            resource_with_link, current=resource_with_link
```
tomeksabala, line 124: image
Update the line with `updated_resource`

tomeksabala (changes requested):

tomeksabala (approved):

PR #147, opened 2021-07-07
ADX-721 add fullpage bulk file uploader

Manoj-nathwani:
# Before
The bulk file uploader was only accessible via the sidebar | A modal would then allow you to select files to upload as new resources
--- | ---
image|image

# After
You can now directly drag-and-drop files onto the page

|With schema |Without schema 
--- | ---
image|image

This will then open the modal just like before with your files selected.

# Changes
- Added new dropzone above resources section of page
- Slightly nicer css on-hover in existing modal dropzone
- Nothing else has changed, there's just been a lot of code reorganising

### DIFF
--- a/ckanext/unaids/react/components/BulkFileUploaderComponent/src/FileUploader.js
+++ b/ckanext/unaids/react/components/BulkFileUploaderComponent/src/FileUploader.js
@@ -1,45 +1,79 @@
 import React from 'react';
 import { useDropzone } from 'react-dropzone'
 
-export default function FileUploader({ maxResourceSize, setPendingFiles }) {
+export const dropzoneTypes = {
+    fullPageDropzone: 'fullPageDropzone',
+    modalDropzone: 'modalDropzone'
+}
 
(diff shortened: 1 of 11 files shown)


ckanext/unaids/react/components/BulkFileUploaderComponent/src/FileUploader.js
```diff
@@ -1,45 +1,79 @@
...
+                                    <h3>{ckan.i18n._('Drop files to upload them')}</h3>
+                                </div>
+                            </div>
+                        }
+                    </>
+                )
+                : (
+                    <>
+                        <input {...getInputProps()} data-testid="BulkFileUploaderInput" />
+                        <p>{ckan.i18n._('Drag & Drop files here to upload')}</p>
```
tomeksabala, line 64: Reword to 'Drop files here to upload them to the dataset'.

tomeksabala (changes requested):

tomeksabala (approved):

PR #154, opened 2021-08-31
Adx 725 partial matching

jonathansberry:
On my holiday I got thinking a bit about the partial filename matching logic as I would love this little feature to work.  Here's what I came up with.

We compare two file names, a and b,  to see if they match:

If a and b are equal, they match.

If a and b both open with the same three or more chars, and close with the same 3 or more chars, and has the same filename extension then we do a regex match on the middle non-matching part. This reg ex will only allow the middle non-matching parts of each filename to consist of the following inserted or swapped:
  - a number
  - a number or the word "copy" in parenthesis
  - a number or the word "copy" after a hyphen

Providing that only one resource matches the new resource name following this logic we will propose the default action of overriding that one resource.   The principle here is to be explicit - it does not matter if things that would ideally be matched don't match, it does matter quite a bit if things that should not match do actually match.  

This code should address the problem of CKAN filename munging I believe (removing parenthesis and inserting hyphens)

The code is tested, appears to work, and I'm pretty happy with it now. 

HOWEVER(!) it has been a long time since I wrote any Javascript, and I am quite sure my style my be outdated and poor.  I welcome any suggested changes from @Manoj-nathwani to ensure this is some good quality Javascript.

### DIFF
--- a/ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
+++ b/ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
@@ -67,19 +67,22 @@ export default function App(props) {
         }
 
         function filenameMatch(filename1, filename2){
-            /* Matches where filenames begin and end with the same three (or more chars) and
-               any middle non-matching parts consist only of the following inserted or swapped:
+            /* Matches where filenames begin and end with the same three (or more chars),
+               have the same file extension, and any middle non-matching parts
+               consist only of the following inserted or swapped:
                - a number
                - a number in parenthesis
                - a number after a hyphen
             */
-            const regexp = /^$|^[\(-]? ?[0-9]*\)?$/;
+
+            const regexp = /^$|^ ?[\(-]? ?([0-9]|copy|Copy)*\)?$/;
             const matches = strDiff(filename1, filename2);
-            const completeMatch = matches[0] == matches[1] && matches[0];
+            const completeMatch = filename1 == filename2;
             const openingMatch = matches[0].length >= 3;
             const closingMatch = matches[1].length >= 3;
+            const extensionMatch = matches[1].includes('.');
             const regexMatch = matches[2].match(regexp) != null && matches[3].match(regexp) != null;
-            const match = completeMatch || openingMatch && closingMatch && regexMatch;
+            const match = completeMatch || openingMatch && closingMatch && extensionMatch && regexMatch;
             return match;
         }
 
--- a/ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
+++ b/ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
@@ -40,38 +40,65 @@ export default function App(props) {
     const [networkError, setNetworkError] = useState(false);
 
     const getDefaultUploadAction = file => {
-        // disabled until we figure out a nicer algorithm
-        // const fuzzySearchResults = new Fuse(
-        //     updateResourcesOptions,
-        //     {
-        //         // https://bit.ly/3BTLnWd
-        //         includeScore: true,
-        //         findAllMatches: false,
-        //         keys: ['fileName'],
-        //         threshold: 0.3
-        //     }
-        // ).search(file.name);
-        // let defaultUploadAction = extraResource;
-        // if (fuzzySearchResults.length === 1) {
-        //     defaultUploadAction = fuzzySearchResults[0].item;
-        // } else if (fuzzySearchResults.length > 1) {
-        //     const firstResult = fuzzySearchResults[0];
-        //     const secondResult = fuzzySearchResults[1];
-        //     const scroreDiff = firstResult.score - secondResult.score;
-        //     if (scroreDiff > 0.05) {
-        //         defaultUploadAction = fuzzySearchResults[0].item;
-        //     }
-        // }
-        // return defaultUploadAction;
-        const exactMatches = updateResourcesOptions.filter(
-            x => x.fileName === file.name
-        );
-        switch (exactMatches.length) {
-            case 1:
-                return exactMatches[0];
-            default:
-                return extraResource;
+
+        function strDiff(str1, str2){
+            // Get all chars that match at the start of both strings
+            var openingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1.charAt(x) == str2.charAt(x)) openingMatch += str1.charAt(x);
+                else break;
+            }
+
+            // Get all chars that match at the end of both strings
+            var str1Reversed = str1.split("").reverse().join("");
+            var str2Reversed = str2.split("").reverse().join("");
+            var endingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1Reversed.charAt(x) == str2Reversed.charAt(x)) endingMatch += str1Reversed.charAt(x);
+                else break;
+            }
+            endingMatch = endingMatch.split("").reverse().join("");
+
+            // Get the middle non-matching part of each string
+            var str1Difference = str1.slice(openingMatch.length, str1.length-endingMatch.length).trim();
+            var str2Difference = str2.slice(openingMatch.length, str2.length-endingMatch.length)
+
+            return [openingMatch, endingMatch, str1Difference, str2Difference];
         }
+
+        function filenameMatch(filename1, filename2){
+            /* Matches where filenames begin and end with the same three (or more chars) and
+               any middle non-matching parts consist only of the following inserted or swapped:
+               - a number
+               - a number in parenthesis
+               - a number after a hyphen
+            */
+            const regexp = /^$|^[\(-]? ?[0-9]*\)?$/;
+            const matches = strDiff(filename1, filename2);
+            const completeMatch = matches[0] == matches[1] && matches[0];
+            const openingMatch = matches[0].length >= 3;
+            const closingMatch = matches[1].length >= 3;
+            const regexMatch = matches[2].match(regexp) != null && matches[3].match(regexp) != null;
+            const match = completeMatch || openingMatch && closingMatch && regexMatch;
+            return match;
+        }
+
+        // Find all matched resources
+        var matchedResources = []
+        for(let x = 0; x<updateResourcesOptions.length; x++){
+            if(filenameMatch(file.name, updateResourcesOptions[x].fileName)){
+                matchedResources.push(updateResourcesOptions[x]);
+            }
+        }
+
+        // Set the default upload action only if there is a single matching resource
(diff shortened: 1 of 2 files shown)


ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+            const openingMatch = matches[0].length >= 3;
+            const closingMatch = matches[1].length >= 3;
+            const extensionMatch = matches[1].includes('.');
+            const regexMatch = matches[2].match(regexp) != null && matches[3].match(regexp) != null;
+            const match = completeMatch || openingMatch && closingMatch && extensionMatch && regexMatch;
+            return match;
+        }
+
+        // Find all matched resources
+        var matchedResources = []
```
Manoj-nathwani, line 90: Replace all `var` with either `const`  or `let`
See https://www.sitepoint.com/es6-let-const/

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+            endingMatch = endingMatch.split("").reverse().join("");
+
+            // Get the middle non-matching part of each string
+            var str1Difference = str1.slice(openingMatch.length, str1.length-endingMatch.length).trim();
+            var str2Difference = str2.slice(openingMatch.length, str2.length-endingMatch.length)
+
+            return [openingMatch, endingMatch, str1Difference, str2Difference];
+        }
+
+        function filenameMatch(filename1, filename2){
```
Manoj-nathwani, line 69: Just an FYI, you don't really use `function` in es6 these days...

Instead of
```
function filenameMatch(filename1, filename2){
    return true
}
```

It would be
```
const filenameMatch(filename1, filename2) => true
```

or 
```
const filenameMatch(filename1, filename2) => {
    return true
}
```

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+                else break;
+            }
+
+            // Get all chars that match at the end of both strings
+            var str1Reversed = str1.split("").reverse().join("");
+            var str2Reversed = str2.split("").reverse().join("");
+            var endingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1Reversed.charAt(x) == str2Reversed.charAt(x)) endingMatch += str1Reversed.charAt(x);
+                else break;
```
Manoj-nathwani, line 58: This is very confusing. It should be:
```
if (str1Reversed.charAt(x) == str2Reversed.charAt(x)){
    endingMatch += str1Reversed.charAt(x);
}else{
    break;
}
```

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
-            default:
-                return extraResource;
+
+        function strDiff(str1, str2){
+            // Get all chars that match at the start of both strings
+            var openingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1.charAt(x) == str2.charAt(x)) openingMatch += str1.charAt(x);
+                else break;
+            }
```
Manoj-nathwani, line 50: This is old school js, es6 has much nicer ways of achieving things like this using `.reduce()`
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

Manoj-nathwani:
This looks great but generally needs syntax updating to a more modern and functional way of writing javascript

Check out
- https://www.codecademy.com/learn/introduction-to-javascript

Things to learn:
- `const` (as variables and functions)
- `Array.prototype.filter()`
- `Array.prototype.map()`
- `Array.prototype.reduce()`

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+            */
+
+            const regexp = /^$|^ ?[\(-]? ?([0-9]|copy|Copy)*\)?$/;
+            const matches = strDiff(filename1, filename2);
+            const completeMatch = filename1 == filename2;
+            const openingMatch = matches[0].length >= 3;
+            const closingMatch = matches[1].length >= 3;
+            const extensionMatch = matches[1].includes('.');
+            const regexMatch = matches[2].match(regexp) != null && matches[3].match(regexp) != null;
+            const match = completeMatch || openingMatch && closingMatch && extensionMatch && regexMatch;
```
Manoj-nathwani, line 85: As a personal note, I always find it confusing chaining so many operators together because you can never be too sure what's being compared with what 😅  So i'd either add a bunch of brackets, declare more variables or do something like this:

````
const match = [
    completeMatch || openingMatch,
    closingMatch,
    extensionMatch,
    regexMatch
].includes(false) === false
```

You may need to run this to confirm the `||` works as expected though

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
-            case 1:
-                return exactMatches[0];
-            default:
-                return extraResource;
+
+        function strDiff(str1, str2){
+            // Get all chars that match at the start of both strings
+            var openingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1.charAt(x) == str2.charAt(x)) openingMatch += str1.charAt(x);
```
Manoj-nathwani, line 48: `==` should be `===`

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+                else break;
+            }
+
+            // Get all chars that match at the end of both strings
+            var str1Reversed = str1.split("").reverse().join("");
+            var str2Reversed = str2.split("").reverse().join("");
+            var endingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1Reversed.charAt(x) == str2Reversed.charAt(x)) endingMatch += str1Reversed.charAt(x);
```
Manoj-nathwani, line 57: `==` should be `===`

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+               have the same file extension, and any middle non-matching parts
+               consist only of the following inserted or swapped:
+               - a number
+               - a number or the word "copy" in parenthesis
+               - a number or the word "copy" after a hyphen
+            */
+
+            const regexp = /^$|^ ?[\(-]? ?([0-9]|copy|Copy)*\)?$/;
+            const matches = strDiff(filename1, filename2);
+            const completeMatch = filename1 == filename2;
```
Manoj-nathwani, line 80: `==` should be `===`

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,68 @@ export default function App(props) {
...
+        var matchedResources = []
+        for(let x = 0; x<updateResourcesOptions.length; x++){
+            if(filenameMatch(file.name, updateResourcesOptions[x].fileName)){
+                matchedResources.push(updateResourcesOptions[x]);
+            }
         }
+
+        // Set the default upload action only if there is a single matching resource
+        var defaultUploadAction = extraResource;
+        if(matchedResources.length == 1){
```
Manoj-nathwani, line 99: `==` should be `===`

Manoj-nathwani (changes requested):

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
Manoj-nathwani: Technically the entire block (declaring a variable, looping and if-statment) can be put into a single `reduce()` statement

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
jonathansberry: AND always before OR, but brackets for clarity is a good idea

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
jonathansberry: Yeah I've read a bit about this and thought a bit.  It's important that I break out of the loop when chars stop matching.  I think I will leave this as it is, but thanks for flagging those map/forEach/reduce functions.

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,74 @@ export default function App(props) {
...
+            let str1Reversed = str1.split("").reverse().join("");
+            let str2Reversed = str2.split("").reverse().join("");
+            let endingMatch = "";
+            for(let x = 0; x < str1.length; x++){
+                if(str1Reversed.charAt(x) === str2Reversed.charAt(x)){
+                    endingMatch += str1Reversed.charAt(x);
+                } else {
+                    break;
+                }
+            }
```
Manoj-nathwani, line 65: This could be a single reduce function
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

ckanext/unaids/react/components/BulkFileUploaderComponent/src/App.js
```diff
@@ -40,38 +40,74 @@ export default function App(props) {
...
+            return match;
+        }
+
+        // Find all matched resources
+        let matchedResources = []
+        for(let x = 0; x<updateResourcesOptions.length; x++){
+            if(filenameMatch(file.name, updateResourcesOptions[x].fileName)){
+                matchedResources.push(updateResourcesOptions[x]);
+            }
         }
```
Manoj-nathwani, line 101: This could be a single reduce function
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

Manoj-nathwani:
FYI: I'm going to have to report this to @tomeksabala for circumventing our 1x PR approval before merging rule 🔥

PR #158, opened 2021-09-28
Adx 714 format guessing

jonathansberry:
This PR will start to autofill the format field when creating a resource. 

Tests for all the server-side logic are included. 

https://user-images.githubusercontent.com/8988344/135090490-6bb60a3c-f753-446d-9130-5ffda43e78ae.mp4


jonathansberry:
Ok tests are failing due to a path in test.ini not being found.  This path is different in local dev and the github actions build. I need to speak to someone about the best way to define them.

ckanext/unaids/assets/format_update.js
```diff
@@ -0,0 +1,44 @@
...
+// Using the action format_guess on the server
+this.ckan.module('format_update', function ($) {
+    return {
+        options: {
+            trigger: 'input[name=url]',
+            filename: 'input[name=url]',
+            guess_action: 'format_guess'
+        },
+        initialize: function () {
+            console.log(this.options);
```
tomeksabala, line 17: remove debug log statment

ckanext/unaids/assets/format_update.js
```diff
@@ -0,0 +1,44 @@
...
+                    function(error){console.error("Error calling api to guess format")}
+                );
+            }else{
+                this._onTrigger();
+            }
+        },
+        _updateFormat: function (json) {
+            if('result' in json && 'format' in json.result){
+                $(this.el).select2("val", json.result.format);
+            }
```
tomeksabala, line 41: Do we want to clear the format select in case of failure. E.g. should we add an `else` block clearing the select?

ckanext/unaids/tests/test_validators.py
```diff
@@ -27,3 +27,20 @@ def test_organisation_id_exists_validator(self):
...
                 {}
             )
+
+    @pytest.mark.parametrize("format,resource_id,url,result", [
+        (None,  None, 'http://adr.local/resource/8919/demo.csv', 'text/csv'),
+        (None,  None, 'http://adr.local/resource/8919/demo.geojson', 'application/geo+json'),
+        (None,  None, 'http://adr.local/resource/8919/demo.pjnz', 'application/pjnz'),
+        ('text/csv',  '8919', 'http://adr.local/resource/89192/demo.xlsx', 'text/csv'),
+        (None,  '8919', 'http://adr.local/resource/8919/demo.csv', None)
+    ])
```
tomeksabala, line 37: I can't understand what exactly this test does. It'd be good to talk this over.

You can't rely on ckan web being up for running the tests. There's a test helper "resource with upload" which can create the test data you need. If it's supposed to be a fake URL it could be named "HTTP://fake.url/resource/1111/demo.csv"

ckanext/unaids/validators.py
```diff
@@ -14,11 +13,9 @@ def if_empty_guess_format(key, data, errors, context):
     # if resource_id then an update
     if (not value or value is df.Missing) and not resource_id:
         url = data.get(key[:-1] + ('url',), '')
-        mimetypes.add_type('application/geo+json', '.geojson')
-        mimetypes.add_type('application/pjnz', '.pjnz')
-        mimetype, encoding = mimetypes.guess_type(url)
-        if mimetype:
-            data[key] = mimetype
+        response = toolkit.get_action('format_guess')({}, {'filename': url})
```
tomeksabala, line 16: IMO it's better to pass the context here since you already have access to it. I think that passing an empty dict is a bad practice.

test.ini
```diff
@@ -10,6 +10,7 @@ port = 5000
 
 [app:main]
 use = config:../ckan/test-core.ini
+ckan.resource_formats = ../../../../adx/ckanext-unaids/ckanext/unaids/resource_formats.json
```
tomeksabala: these can be done with `%(here)s` *.ini keyword. I've pushed a fix.

tomeksabala (changes requested):
I would be good to talk over this changes in person to help me understand some details.

jonathansberry:
Ok so tests are passing locally but failing in github actions. 

As I understand it:

- Python `mimetypes` includes an inbuilt mapping of file extension to mimetypes.
- This internal mapping is augmented by local files on the system - in many linux distributions this includes `/etc/mime.types` 
- Due to this dependency on local files the python mimetypes module is not matching xlsx to any particular mimetype in the github actions. 

I suggest that this is treated as a seperate issue if the matter arises as an actual problem and I simply fix the tests by switching from a .xlsx filename to a .xls filename

tomeksabala (approved):

PR #164, opened 2021-11-28
NAV-769 Added blob storage fields validator for resource update/create

tomeksabala:


jonathansberry (approved):
Great solution I think, and solid testing. 

Have we checked what actually happens in the UI upon clicking "finish" for a bad upload? I'm just not sure how these validation errors will be flagged to the user in practise?

PR #169, opened 2022-01-10
ADX-792 Default giftless metadata included at page load. 

jonathansberry:
Problem: users have edited and saved metadata in less than the time it takes for the FileInputComponent to load.  This has meant that the giftless metadata, which is added to the form submission by the file input component, was lost. 

Possible solutions:

- Delay form submission until the file input component had properly loaded
- Ensure that the relevant information is supplied as part of the page HTML, so that the form can be used even if the FileInputComponent entirely fails to load.

I have pursued the second option, as this is how we are approaching the matter everywhere else in the CKAN UI; the HTML responsible for the giftless metadata that is created/updated by the FileInputComponent react code is now initialised by the server upon page load. I've used Chrome request throttling to test it works in my UI.  Without a proper UI testing framework it is hard to write a regression test.

### DIFF
--- a/ckanext/unaids/tests/test_logic.py
+++ b/ckanext/unaids/tests/test_logic.py
@@ -10,6 +10,8 @@ from ckanext.unaids import logic
     [
         ("fjelltopp/my-dataset", "acbac3b78f9ace071ca3a79f23fc788a1b7ee9dc547becc6404dbb1f58afff79", 100, True),
         (None, None, None, True),
+        ("", "", "", True),
+        (None, "", "", False),
         ("fjelltopp/my-dataset", "acbac3b78f9ace071ca3a79f23fc788a1b7ee9dc547becc6404dbb1f58afff79", None, False),
         (None, None, 100, False),
         ("fjelltopp/my-dataset", "", 100, False),
@@ -22,6 +24,8 @@ from ckanext.unaids import logic
     ], ids=[
         "validates if all upload fields correct",
         "validates if no upload fields present",
+        "validates if all fields are present but empty",
+        "fails if some fields missing and others set to empty string",
         "fails if size None",
         "fails if only size present",
         "fails if sha256 empty string",
--- a/ckanext/unaids/theme/templates/scheming/form_snippets/upload.html
+++ b/ckanext/unaids/theme/templates/scheming/form_snippets/upload.html
@@ -19,6 +19,13 @@
             <h3 class="text-muted">
                 <span><i class="fa fa-spinner fa-spin"></i> {{ _('Loading') }}</span>
             </h3>
+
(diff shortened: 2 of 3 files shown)


ckanext/unaids/theme/templates/scheming/form_snippets/upload.html
```diff
@@ -37,8 +37,15 @@ <h3 class="text-muted">
...
             <h3><i class="fa fa-spinner fa-spin"></i> {{ _('Loading') }}</h3>
-        </div>        
+        </div>
     {% endif %}
+
+    <input name="url" data-testid="url" type="hidden" value="{{ data.url if data else '' }}">
+    <input name="url_type" data-testid="url_type" type="hidden" value="{{ data.url_type if data else '' }}">
+    <input name="lfs_prefix" data-testid="lfs_prefix" type="hidden" value="{{ data.lfs_prefix if data else '' }}">
+    <input name="sha256" data-testid="sha256" type="hidden" value="{{ data.sha256 if data else '' }}">
+    <input name="size" data-testid="size" type="hidden" value="{{ data.size if data else '' }}">
```
tomeksabala, line 47: if we always send empty `size`, `sha256` and `lfs_prefix` will not allow users to create empty resources (without resource fiels).

Changing it to:
```
{% if data.lfs_prefix %}
<input name="lfs_prefix" data-testid="lfs_prefix" type="hidden" value="{{ data.lfs_prefix}}">
{% endif %}
```
only partially resolves the problem as it'll still not allow users to **remove** the file from a resource in an already existing resource.
I don't know how to enable support for removing the resource files.

tomeksabala (changes requested):

ckanext/unaids/theme/templates/scheming/form_snippets/upload.html
jonathansberry: We have spotted that this is actually a problem with the validator logic.  It should be possible for giftless metadata to be all empty strings. I have updated the PR to include the fix to the validator logic.

ckanext/unaids/tests/test_logic.py
```diff
@@ -22,6 +24,8 @@
     ], ids=[
         "validates if all upload fields correct",
         "validates if no upload fields present",
+        "validates if all fields are present but empty",
+        "fails if some fields missing and others set to empty string",
```
jonathansberry, line 28: I am not sure we really want this? i.e. If some fields are missing and all others present but empty, should this pass validation or not? I have not allowed this here since it is an edge case that would involve a much more substantial change to the original validator logic.

tomeksabala: I think it's ok behaviour which disables partial input (you should either provide all or none blob storage metadata).

tomeksabala (approved):

PR #171, opened 2022-01-14
ADX-279 Add API action user_show_me

jonathansberry:
Adds a new API action that simply returns the current user obj.  Can be used to verify if an API key is valid. Tests included.

Is there a security concern around this endpoint?  Should we throttle the number of times someone can call it?

### DIFF
--- a/ckanext/unaids/actions.py
+++ b/ckanext/unaids/actions.py
@@ -170,3 +170,12 @@ def format_guess(context, data_dict):
         return {'mimetype': mimetype, 'format': format}
     else:
         return {'mimetype': None, 'format': None}
+
+
+@logic.side_effect_free
+def user_show_me(context, resource_dict):
+    auth_user_obj = context.get('auth_user_obj')
+    if auth_user_obj:
+        return auth_user_obj.as_dict()
+    else:
+        return {}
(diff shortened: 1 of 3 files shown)


jonathansberry:
> The feature looks and work alright for valid auth. However for incorrect apikey it returns:
> 
> ```
> {
>     "help": "http://adr.local/api/3/action/help_show?name=user_show_me",
>     "success": true,
>     "result": {}
> }
> ```
> 
> I think that this response format might be consusing. We have a "success: true" response with no result. How about:
> 
> 1. return 404
> 
> ```
> {
>     "help": "http://adr.local/api/3/action/help_show?name=user_show",
>     "success": false,
>     "error": {
>         "message": "Not found",
>         "__type": "Not Found Error"
>     }
> }
> ```
> 
> 2. return 403 forbidden (my recommendation) e.g.:
> 
> ```
> {
>     "help": "http://adr.local/api/3/action/help_show?name=send_email_notifications",
>     "success": false,
>     "error": {
>         "message": "Access denied: Action send_email_notifications requires an authenticated user",
>         "__type": "Authorization Error"
>     }
> }
> ```

My proposed design is consistent with the user_organization_list endpoint, which returns an empty list if the user hasn't logged in.   I've just spent some time reviewing the API to see if there are any other similar endpoints.  I can't see any - all other endpoints require a user_id to be manually specified. I think in light of this, I agree with this proposed change, I will implement and push.

tomeksabala (approved):

PR #175, opened 2022-02-17
ADX-470 data dictionaries

jonathansberry:
This PR introduces auto-population of data dictionaries from Frictionless Data table schemas used in data validation.

Current logic is to update the field's 'label' and 'description' in the data dictionary only if a corresponding field exists in the validation table schema and if no pre-existing label or description exists in the data dictionary for the field.

The logic fails silently if you try to auto-populate the data dictionary for a resource that has no associated table schema.  It will throw errors for all other failure cases.  In the "after_upload" hook we want complete silent failure though, since the task runs as a background. I have therefore decided to catch all exceptions and log them, but not block code execution, as this causes significant problems down stream. 

For convenience an action has been included to run the logic.  This will be especially useful in the migration script. 

To do:

- Test in development
- Run migration script for existing resources tested in development
- Deploy to production

### DIFF
--- a/ckanext/unaids/logic.py
+++ b/ckanext/unaids/logic.py
@@ -1,5 +1,6 @@
 from ckan import model
 from ckan.plugins import toolkit
+from ckanext.validation.helpers import validation_load_json_schema
 
 
 def validate_resource_upload_fields(context, resource_dict):
@@ -35,3 +36,38 @@ def update_filename_in_resource_url(resource):
         new_url_segments = url_segments[:-1] + [filename]
         resource['url'] = '/'.join(new_url_segments)
     return resource
+
+
+def auto_populate_data_dictionary(context, resource_dict, dataset_dict):
+    table_schema = resource_dict.get('schema')
+
+    if not table_schema:
+        return
+
+    table_schema_dict = validation_load_json_schema(table_schema)
+
+    if not table_schema_dict or not table_schema_dict.get('fields'):
+        return
+
+    field_schemas = {field['name']: field for field in table_schema_dict['fields']}
+    fields = toolkit.get_action(u'datastore_search')(
+        context, {u'resource_id': resource_dict['id']}
+    )[u'fields']
+
+    fields = fields[1:]  # Hack: to get rid of _id field.  But this should be got rid of higher up stream.
+
+    for field in fields:
+        field_id = field[u'id']
+        if field_id in field_schemas and not field.get(u'info'):
+            field[u'info'] = {
+                u'label': field_schemas[field_id].get(u'title', field_id),
+                u'notes': field_schemas[field_id].get(u'description', field_id)
+            }
+
+    toolkit.get_action(u'datastore_create')(
+        context, {
+            u'resource_id': resource_dict[u'id'],
+            u'force': True,
+            u'fields': fields
+        }
+    )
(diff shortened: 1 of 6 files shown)


ckanext/unaids/logic.py
```diff
@@ -35,3 +36,49 @@ def update_filename_in_resource_url(resource):
         new_url_segments = url_segments[:-1] + [filename]
         resource['url'] = '/'.join(new_url_segments)
     return resource
+
+
+def auto_populate_data_dictionary(context, resource_dict):
+    table_schema_name = resource_dict.get('schema')
+
+    if not table_schema_name:
+        return
```
tomeksabala, line 45: In the PR description, you're saying
> The logic fails silently if you try to auto-populate the data dictionary for a resource that has no associated table schema.

I wouldn't call this behaviour "fail silently" but rather bypassed. Did I miss  something?

This `return None` behaviour here might be confusing for API users. How about using `toolkit.get_or_bust(resource_dict, 'schema')` here and handling running the auto population only for schemed resources in the `after_upload` handler:
```
    def after_upload(self, context, resource_dict, dataset_dict):
       if 'schema' in resource_dict:
            try:
                logic.auto_populate_data_dictionary(context, resource_dict)
            except Exception as e:
```

That way your intent is much more explicit and `auto_populate_data_dictionary` errors more friendly.

ckanext/unaids/logic.py
```diff
@@ -35,3 +36,49 @@ def update_filename_in_resource_url(resource):
...
+
+    table_schema = validation_load_json_schema(table_schema_name)
+
+    if not table_schema:
+        raise toolkit.ObjectNotFound(
+            'Resource table schema "{}" does not exist'.format(table_schema)
+        )
+
+    field_schemas = {field['name']: field for field in table_schema.get('fields')}
```
tomeksabala, line 54:
```
{field['name']: field for field in table_schema.get('fields', [])}
```
default value empty list instead of `None`. Otherwise you'll get:
```
TypeError: 'NoneType' object is not iterable
```

tomeksabala (changes requested):

ckanext/unaids/tests/test_actions.py
```diff
@@ -192,3 +191,22 @@ def test_user(self):
...
+@pytest.mark.ckan_config('ckan.plugins', 'unaids')
+@pytest.mark.usefixtures('with_plugins')
+class TestPopulateDataDictionary(object):
+
+    def test_expected_use(self, mocker):
+        dataset = factories.Dataset()
+        resource = factories.Resource(
+            package_id=dataset['id'],
+            schema='test_schema',
+            url='some.data'
```
jonathansberry, line 205: @tomeksabala I have had to add this line in because my test broke after rebasing against development for no reason that I could understand.  Some further inspection suggests perhaps it is a bug in your filename munging solution here: https://github.com/fjelltopp/ckanext-unaids/blob/a0c6b9af5d5afd0e6551600ca114cb4be12282c9/ckanext/unaids/logic.py#L35

Basically the default url value supplied by the factory is `http://link.some.data` I think your code keeps adding extra copies of `http://` to the start of the strin i.e `http://http://http://link.some.value` depending how often it is hit.  Can you investigate?  To reproduce just remove this line and re run this test.

tomeksabala: ok, I'll have a look into that.

tomeksabala (approved):

ckanext/unaids/tests/test_actions.py
tomeksabala: I have a fix for that but let's merge it in a sep PR after this one.

PR #177, opened 2022-03-01
ADX-767-fix-upload-button-not-being-disabled

Manoj-nathwani:
- https://fjelltopp.atlassian.net/browse/ADX-767
- Video showing fix: https://youtu.be/SyfIjIxEH0c

### DIFF
--- a/ckanext/unaids/react/components/BulkFileUploaderComponent/src/ModalBody.js
+++ b/ckanext/unaids/react/components/BulkFileUploaderComponent/src/ModalBody.js
@@ -185,9 +185,9 @@ export default function ModalBody({
             .map(file => uploadActionAlreadyTaken(file.uploadAction))
             .filter(x => x).length === 0;
         const enableButton = [
-            !thereArePendingFiles,
+            thereArePendingFiles,
             !uploadInProgress,
-            !noDuplicateUploadActionsSet
+            noDuplicateUploadActionsSet
         ].every(Boolean);
         return (
             <button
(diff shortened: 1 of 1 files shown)


ckanext/unaids/react/components/BulkFileUploaderComponent/src/ModalBody.js
```diff
@@ -186,6 +186,7 @@ export default function ModalBody({
             .filter(x => x).length === 0;
         const enableButton = ![
             thereArePendingFiles,
+            !uploadInProgress,
```
tomeksabala, line 189:
```        
const enableButton = ![
            thereArePendingFiles,
            !uploadInProgress,
            noDuplicateUploadActionsSet
        ].includes(false);
```
This conditional is quite complicated. You could simplify it a bit with:
```
const enableButton = [
            !thereArePendingFiles,
            uploadInProgress,
            !noDuplicateUploadActionsSet
        ].every(Boolean);
```
More info about `Array.every(Boolean)` https://stackoverflow.com/questions/53897673/check-if-all-values-in-array-are-true-then-return-a-true-boolean-statement-jav

This new aproach is much easier to read in English and understand when exactly we want the button to be disabled.
In other words we're chaning the logic expression of `!(!A or !B or !C)` into `(A and B and C)`.

tomeksabala (changes requested):

ckanext/unaids/react/components/BulkFileUploaderComponent/src/ModalBody.js
Manoj-nathwani: Yeah, this is good! I didn't know of `.every()`

tomeksabala (changes requested):
tests are failing

tomeksabala (approved):

PR #186, opened 2022-03-18
ADX-561-cookie-policy-popup

Manoj-nathwani:
https://fjelltopp.atlassian.net/browse/ADX-561

Added a popup banner in the bottom right of the page to confirm you understand that cookies are in use.

image


Manoj-nathwani:
Marking as stale

PR #190, opened 2022-03-28
ADX-749-datasettransfer-not-working-for-collaborator

Manoj-nathwani:
https://fjelltopp.atlassian.net/browse/ADX-749

### DIFF
--- a/ckanext/unaids/blueprints/unaids_dataset_transfer.py
+++ b/ckanext/unaids/blueprints/unaids_dataset_transfer.py
@@ -14,11 +14,6 @@ unaids_dataset_transfer = Blueprint(
 )
 
 
-def _admin_user():
-    context = {'model': model, 'session': model.Session, 'ignore_auth': True}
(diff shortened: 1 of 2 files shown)


ckanext/unaids/blueprints/unaids_dataset_transfer.py
```diff
@@ -31,9 +36,9 @@ def process_dataset_transfer(dataset_id):
         'org_to_allow_transfer_to': None
     })
     toolkit.get_action('package_update')({
+        'user': _admin_user()['name'],
```
tomeksabala, line 39: I have described the reason why it's needed here: https://github.com/ckan/ckan/issues/6779
An easier workaround to this bug is to pass empty `'user'` value like:
```
toolkit.get_action('package_update')({
        'user': ''
        'model': model,
        'session': model.Session,
      'ignore_auth': True
```

tomeksabala (changes requested):

tomeksabala (approved):

PR #193, opened 2022-04-01
ADX-744-hide-featured-banner-on-mobile

Manoj-nathwani:
https://fjelltopp.atlassian.net/browse/ADX-744
image

### DIFF
--- a/ckanext/unaids/theme/templates/home/layout1.html
+++ b/ckanext/unaids/theme/templates/home/layout1.html
@@ -1,10 +1,10 @@
 <div role="main" class="hero">
   <div class="container">
     <div class="row row1">
-      <div class="col-md-6 col1">
+      <div class="col-md-6 col1 hidden-xs">
         {% block promoted %}
         <div class="module-content box">
-          <section class="featured media-overlay hidden-xs">
+          <section class="featured media-overlay">
             <center>
             <!--<a class="media-image" href="/about">
               <!--<img class="img-responsive" src="{{ h.url_for_static('/aim.png') }}" alt="aim" width="420" height="220" /> -->


ckanext/unaids/theme/templates/home/layout1.html
```diff
@@ -1,10 +1,10 @@
 <div role="main" class="hero">
   <div class="container">
     <div class="row row1">
-      <div class="col-md-6 col1">
+      <div class="col-md-6 col1 hidden-xs">
```
tomeksabala, line 4: Hiding this div above `promoted` block looks a little off. Have you compared the layout1 template we're extending with the original core CKAN and potentially see how it's resolved there?

tomeksabala (changes requested):

ckanext/unaids/theme/templates/home/layout1.html
jonathansberry: I think using hidden-xs is not a bad solution, but I just couple of points for consideration:

1. Have you considered whether it is easy to actually show the content in mobile view? It's worth giving maybe 10 minutes thought to - fiddling in the browsers inspector console - if after 10 minutes you don't have an easy fix just don't bother. 
2. We need to ensure that the screen size at which the div content disappears exactly aligns with the screen size at which the div itself disappears.  The best way to do this would be to not hide the content at all, but just have one place where the entire div is hidden.  i.e. check which bit of the CSS is already doing the hiding, and ensure that the entire div is hidden with a single CSS rule.

Manoj-nathwani: @jonathansberry Yea I think we can keep it on mobile. see:
- https://fjelltopp.atlassian.net/browse/ADX-812
- https://github.com/fjelltopp/ckanext-unaids/pull/195

Manoj-nathwani:
This PR isn't needed as https://github.com/fjelltopp/ckanext-unaids/pull/195 was implemented

PR #202, opened 2022-04-06
Remove unnecessary console.log

Manoj-nathwani:
Looks like this sneakily made it out into prod 👀


tomeksabala (changes requested):
Good catch. Please change the base branch  to `development`

jonathansberry (approved):

PR #212, opened 2022-04-19
ADX-810-cleanup-datapreview-css

Manoj-nathwani:
https://fjelltopp.atlassian.net/browse/ADX-810

## Before
image

## After
image

### DIFF
--- a/ckanext/unaids/theme/templates/base.html
+++ b/ckanext/unaids/theme/templates/base.html
@@ -1,11 +1,7 @@
 {% ckan_extends %}
 
-{% block styles %}
-  {{ super() }}
-  {% asset 'ckanext-unaids/UnaidsCSS' %}
-{% endblock %}
-
 {% block head_extras %}
+  {% asset 'ckanext-unaids/UnaidsCSS' %}
   {{ super() }}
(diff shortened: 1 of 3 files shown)


ckanext/unaids/theme/templates/base.html
```diff
@@ -1,11 +1,7 @@
 {% ckan_extends %}
 
-{% block styles %}
-  {{ super() }}
-  {% asset 'ckanext-unaids/UnaidsCSS' %}
-{% endblock %}
-
 {% block head_extras %}
+  {% asset 'ckanext-unaids/UnaidsCSS' %}
   {{ super() }}
```
Manoj-nathwani, line 5: The reason I had to do this was that `recline_view.html` overwrites the `{% block styles %}` block in the template
https://github.com/ckan/ckan/blob/master/ckanext/reclineview/theme/templates/recline_view.html

So we need to make sure our css fixes are still applied.

An alternative approach would be to make a specific `recline_view_fixes.css` and apply it to `{% block head_extras %}` but I think applying the whole of `/unaids/assets/custom.css` is still ok

tomeksabala (approved):

PR #225, opened 2022-08-09
FIP-14 Test GH Superlinter

fulior:


tomeksabala:
The new linter should honor our flake8 configuration (earlier passed as cli arguments).

PR #227, opened 2022-08-17
ADX-805 Remove auto-generation of resources

jonathansberry:
This PR removes the feature that auto-generates resources.  This is because:

- The feature is very old and no longer actively under use
- Has been broken for some time without any users raising concerns
- Creates an amalgamation of existing data purely for convenience, we are not deleting or removing any data from the system.
- Was always a rapidly built prototype solution a specific problem faced whilst on mission in Johannesburg.  We have gone in a very different direction now.
- The code now requires maintenance and bug fixing which is just blocking further important development.


jonathansberry:
Test now pass locally.

PR #235, opened 2022-09-14
ADX 841 Add user affiliation

ChasNelson1990:
* Adds 'user affiliation' and 'job title' fields to the user profile
* When existing users next log in they are redirected to the user profile page with a message asking them to update their details
* New users cannot be created without these details
* User profiles cannot be edited with these fields empty
* Tests have been added for each action and the blueprint

Additional: I discovered when running the tests that this breaks a lot of `ckanext-unaids` tests because they user `factories.User()`, which doesn't include these new fields and so falls flat on its face. I have fixed this by adding a new wrapper function in `tests` called `user_factory_with_affiliation()`. However, there is still a clash with `ckanext-versions` 
the fixtures all use `factories.User()` and so the tests that load that plugin fail - is there a better way to do this? i.e. can I actually overload the factory rather than wrapping it?


ckanext/unaids/tests/__init__.py
```diff
@@ -43,3 +43,13 @@ def create_dataset_with_releases(user, number_of_releases=5):
...
+
+
+def user_factory_with_affiliation(**kwargs):
+    if "job_title" not in kwargs.keys():
+        kwargs["job_title"] = "Data Scientist"
+    if "affiliation" not in kwargs.keys():
+        kwargs["affiliation"] = "Data Fjelltopp"
+    return factories.User(
+        **kwargs
+    )
```
jonathansberry, line 55: See my alternative suggestion involving switching the feature on and off, in order to get around this issue.

ckanext/unaids/tests/test_actions.py
```diff
@@ -209,3 +208,59 @@ def test_expected_use(self, mocker):
...
+        )
+        assert user.get('job_title', False) == 'Data Scientist'
+        assert user.get('affiliation', False) == 'Fjelltopp'
+
+    @pytest.mark.parametrize('job_title, affiliation', [
+        ('Data Scientist', None),
+        (None, 'Fjelltopp'),
+        ('Data Scientist', ''),
+        ('', 'Fjelltopp'),
+    ])
```
jonathansberry, line 230: NOICE :+1:

jonathansberry (changes requested):
Great work Chas.  The tests are simple and decent.  I've suggested an alternative solution to wrapping the User() factory everywhere. Also a few other suggestions around ckan style and expression.  Let me know what you think when you are back from vacation.

ckanext/unaids/blueprints/user_affiliation.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # the default is dashboard.index
+    # however, and ".index"es will not get caught at, for example, "/dashboard/"
+    # so we remove any ending ".index"
+    if configured_route.endswith('.index'):
+        configured_route = configured_route[:-6]
+
+    # add the leading and trailing slashes
+    # and convert any internal periods to slashes
+    return '/' + configured_route.replace('.', '/') + '/'
```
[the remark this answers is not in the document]
ChasNelson1990: This broke my tests because `url_for` (and the ckan.lib.helpers version) needs to be run within a flask application context and seem not to have this during tests. So I have left as is.

ckanext/unaids/tests/test_blueprints.py
```diff
@@ -17,8 +17,14 @@ def test_validate_user_profile_blueprint_intercepts(self, app):
         ckan.plugins.unload('unaids')
         user = User()
         ckan.plugins.load('unaids')
-        url_after_login = get_route_to_intercept()
-        user_response = app.get(url_after_login, headers={
-            'Authorization': user['apikey']
-        }, follow_redirects=False)
+        # url_after_login = get_route_to_intercept()
```
jonathansberry, line 20: Can we just get rid of the line?

jonathansberry (changes requested):
We're there! Or basically there! Just spotted a few commented out lines of code.  We should just delete them.

tomeksabala:
I've just noticed that the demo data loader configuration should also be updated with the 2 new user fields: affiliation and job title.
Have a look at: `adx_develop/demo_data/users.json`

ChasNelson1990:
> I've just noticed that the demo data loader configuration should also be updated with the 2 new user fields: affiliation and job title. Have a look at: `adx_develop/demo_data/users.json`

Whoops, yes, I did that but forgot to push adx_develop at all. Ta.

jonathansberry (approved):

PR #238, opened 2022-10-18
ADX841 Custom edit user form for job title and affiliation

ChasNelson1990:
Here we add a new blueprint that is essentially a custom, truncated to focus on our new custom fields.

The new view for this is heavily based on ckan's EditView but simplified for our scenario.

TODO:
- [ ] Update the example job title and affiliation


ChasNelson1990:
Looking at this, I don't know if we want to factor the View out to it's own file as it's no longer purely about validation, as it now does edit too?

ChasNelson1990:
Closing as the client didn't want this feature

PR #241, opened 2022-11-04
ADX-849 Resource forking UI

ChasNelson1990:
- Removed a redundant npm lock file
- Updated how we use reactDOM
- Added the new ResourceForker component with v0 CSS based on bootstrap and existing CSS classes
- Add an "Imported from:" line to resources that are forks
- Make the styling more in line with our mock ups
- Add checks for authorisation
- Prevent self selecting
- Add badges to show syn status
- Add button for manual re-sync

These changes are dependent on: fjelltopp/ckanext-fork#11

postponed:

- [ ] Add activity stream line for forking -  we have decided to pin this, hoping for access to the right ckan lib code in 2.10 when activities is spun out as a plugin
- [ ] Add GitHub action for running Jest tests (@ChasNelson1990 has one he can copy over)
- [ ] Move to JSX file extension (Javascript with React)
- [ ] Add ESLint (& prettier?) and lint to bring all components up to industry standards


ckanext/unaids/react/components/FileInputComponent/src/App.js
```diff
@@ -27,105 +40,142 @@ export default function App({ loadingHtml, maxResourceSize, lfsServer, orgId, da
...
                         lfs_prefix: null,
                         sha256: null,
                         size: null,
                         // url field is handled by input field in UI
-                    }
+                        fork_resource: null
+                    };
+                case 'resource':
+                    if (fileFormatField) fileFormatField.value = metadata.format;
+                    // setUploadFileName(metadata.filename);
```
tomeksabala, line 57: stale comment?

ckanext/unaids/react/components/FileInputComponent/src/App.js
```diff
@@ -16,9 +23,15 @@ export default function App({ loadingHtml, maxResourceSize, lfsServer, orgId, da
...
     const [useEffectCompleted, setUseEffectCompleted] = useState(false);
+    const [selectedResource, setSelectedResource] = useState({
+        resource: null,
+        dataset: null,
+    });
 
     const setHiddenInputs = (newUploadMode, metadata) => {
         setUploadMode(newUploadMode);
+        let fileFormatField =
+                        document.getElementById('field-format');
```
tomeksabala, line 34: we have a line size limit set to 127. No need to break it here.

ckanext/unaids/react/components/FileInputComponent/src/App.js
```diff
@@ -27,105 +40,142 @@ export default function App({ loadingHtml, maxResourceSize, lfsServer, orgId, da
...
+        } else if (uploadMode === 'resource') {
+            return (
+                <ResourceForker
+                    selectedResource={selectedResource}
+                    setSelectedResource={setSelectedResource}
+                    setHiddenInputs={setHiddenInputs}
+                />
+            );
+        } else {
+            // e.g. [undefined, null, "file"].includes(uploadMode)
```
tomeksabala, line 158: I don't understand this comment. What's it about?

tomeksabala (changes requested):

ckanext/unaids/react/components/FileInputComponent/src/App.js
ChasNelson1990: Where is this set for JS/TS?

ckanext/unaids/react/components/FileInputComponent/src/App.js
ChasNelson1990: The code was originally written to do `if <that line>; else...` and I flipped it around so I kept that more explicit statement in so that future us can understand what that is actually trying to catch. Perhaps the correct solution is to make this branch an `elseif` and have an `else` that throws an exception.
(Philosophically I would actually say that the correct solution should be to use an enum but you're using JS not TS... one for a broader discussion I'm sure :-) )

ckanext/unaids/react/components/FileInputComponent/src/App.js
ChasNelson1990: Actually... is this even set somewhere for Python inside the repo? I don't see a .flake8 or a [flake8] section in the pyproject.toml?

ckanext/unaids/react/components/FileInputComponent/src/App.js
ChasNelson1990: my bad

ckanext/unaids/react/components/FileInputComponent/src/App.js
tomeksabala: we have the limit hardcoded in flake8 command inside github actions config.

ckanext/unaids/react/components/FileInputComponent/src/App.js
```diff
@@ -155,7 +153,8 @@ export default function App({
                 />
             );
         } else {
-            // e.g. [undefined, null, "file"].includes(uploadMode)
+            // essentially this is catchin any of [undefined, null, "file"]
+            // and anything else will also  be treated as this general case
```
tomeksabala, line 157: I think this comment creates more confusion then help. Either we actually care about value of uploadMode here (then you should put the condition in else if) or we don't care and any value is treated as general case.

If a developer would like to get all possible values of uploadMode then he should look for usage of setUploadMode/setHiddenInputs in the code. Such comment is extremely prone to become outdated and in sucha  case be even more confusing.

tomeksabala (changes requested):

jonathansberry (approved):
Fantastic.  It's looking really good now.  I have added some formatting changes, including the resource icons. 

There are still some small issues that need addressing.  But I suggest that these remaining issues do not block deployment. The feature works well enough for the TOT and creating help videos. Two things I think we need to revisit in order of priority (can you create Jira issues @ChasNelson1990):

1. You currently *remove* the current resource from the search results.  This creates strange behaviour, such as a dataset returned in the search results that apparently has no search matches (because the one search match was on the current resource). I think a simpler/better approach is to "disable" the current resource in the search results in a similar manner to the way you disable resource the user doesn't have access to view.  
2. There's something a little odd going on with the last modified date on the resource-details-tile.  When hitting "syncronise data" the last modified date updates - which is great! but it updates to (i think) a timestamp that was fetched upon page load, rather than upon clicking "syncronise now". This will create strange behaviour in edge cases.  It is a bug worth fixing soon, but not a bug worth blocking deployment.

PR #246, opened 2022-11-17
ADX-883 Remove VMMC button

jonathansberry:
Small PR to remove VMMC button from the home page.

image


tomeksabala (changes requested):
What about removal of VMMC from the dataset selector:
`ckanext/unaids/theme/templates/snippets/dataset_selector.html` ?

tomeksabala (approved):

PR #257, opened 2022-11-29
ADX-895 Enable job affiliation & org name import from Auth0 with ADR SSO

ChasNelson1990:
Two commits:

1. adds the ability to read the profile info from the SAML data (requires Auth0 changes)
2. removes all the unused code for the old ckan-only way

PR #258, opened 2022-11-29
ADX-897 Some log in links are not pointing out to SSO login and display error message instead

ChasNelson1990:
Basically, I have stolen what ckanext-saml2auth does to create the "This page doesn't exist now you have SSO" catch but instead of a kind of annoying message I just redirect to the saml2 login screen.

What do we think of this approach?

TODO:
- [ ] tests

P.S. I have built this off of `development` so you may like to take #257 into account whilst reviewing.


ckanext/unaids/blueprints/login_register_catch.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def disable_default_login():
+    """
+    Override and redirect default Login route
+    """
+    return toolkit.redirect_to("/user/saml2login")
+
+
+def disable_default_register():
```
tomeksabala, line 17: I think that right now it's best to leave only one function here used for both blueprint rules.

tomeksabala (dismissed):
please rebase this branch on the other pieces of work

tomeksabala (approved):

PR #264, opened 2022-12-08
Add hidden passwords field for CKAN to not crash when updating About text or upload a new Profile Picture

A-Souhei:
Minimal changes in the template **edit_user_form.html**, I added back 3 old passwords inputs but as hidden;
Did not use `form.input() `due to it adding a label tag before the input.

### DIFF
--- a/ckanext/unaids/theme/templates/user/edit_user_form.html
+++ b/ckanext/unaids/theme/templates/user/edit_user_form.html
@@ -34,6 +34,12 @@
 
   </fieldset>
 
+   <div>
+      <input type="hidden" name="old_password"/>
+      <input type="hidden" name="password1"/>
+      <input type="hidden" name="password2"/>
+  </div>
+
   <div class="form-actions">
     {% block delete_button %}
       {% if h.check_access('user_delete', {'id': data.id})  %}


ChasNelson1990 (commented):
Sensible change and generally looks fine but I have a minor question - do we need all three? Does ckan not always check `password1` first and so if `password1` is always `None` we don't need `password2` or `old_password`?

Also, do we want a regression test for this?

A-Souhei:
@ChasNelson1990 these inputs are used in **adx_develop/submodules/ckan/ckan/views/user.py**

```
        if (data_dict[u'password1']
                and data_dict[u'password2']) or email_changed:
            identity = {
                u'login': g.user,
                u'password': data_dict[u'old_password']
            }
```

Another solution would be to find a way to update the function in the CKAN core without actually changing the CKAN core, but this depends on the review since it could be more difficult.

ChasNelson1990:
> ```
>         if (data_dict[u'password1']
>                 and data_dict[u'password2']) or email_changed:
> ```

So Python should fast fail on an `and`, i.e. is `data_dict[u'password1']` is `False`y then it shouldn't bother looking for `password2`... as I say, it's a pretty minor thing.

A-Souhei:
It always crashes for me, could you check on your side when you have time?

ChasNelson1990:
> It always crashes for me, could you check on your side when you have time?

Well if it crashes then I am wrong lol! So please ignore me then.

I will rereview

ChasNelson1990 (approved):
LGTM

PR #306, opened 2023-09-29
Adx 905 job title and affiliation in new membership request mails

A-Souhei:
## Description

Me, @tomeksabala , have take over this PR and continued working on it. Here are my description of the changes:

1. I've made refactor to encapsulate all logic around custom user profile fields (job title & affiliation) in a single package `custom_user_profile`
2. The email templates are defined and later used in email sending in the ckanext-ytp-request. However, custom user profile fields are defined in ckanext-unaids. That's why I needed to duplicate a piece of logic of ytp-request in `custom_user_profile/action.py`.
This change also requered (due to a @chained_action of ytp-request in ckanext-unaids/plugin.py) to add the `ytp_request` plugin in all `@ckan.config('plugins' [...]` in the test code 

## Checklist


A-Souhei:
The test is unfinished even though Chas helped solve a part of it, I will find some time to continue

ckanext/unaids/custom_user_profile/actions.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # TODO: Is there a way to get language associated to all admins. User table there is nothing as such stored
+    locale = get_safe_locale()
+
+    member.state = 'pending'
+    member.capacity = role
+    member.group = group
+
+    logging.warning("Member's Group ID: " + str(type(member.group_id)))
+    logging.warning(repr(member))
```
jonathansberry, line 101: Are these really valid warning logs?  Looks a bit like debugging print statements to me.

ckanext/unaids/custom_user_profile/actions.py
```diff
@@ -0,0 +1,157 @@
...
+    try:
+        mail_user(admin, subject, message)
+    except Exception:
+        log.exception("Mail could not be sent")
+
+
+def _create_member_request(context, data_dict):
+    """
+    Whole method copied over from ckanext-ytp-request plugin
+    Helper to create member request
```
jonathansberry, line 54: I believe since it ytp_request is our fork and no longer actively maintained upstream, we can just edit ytp_request code in place, instead of overriding from ckanext-unaids? This approach of overriding from ckanext-unaids is creating significant amounts of duplicate code?

ckanext/unaids/custom_user_profile/actions.py
```diff
@@ -0,0 +1,157 @@
...
+from ckan.plugins.toolkit import config
+from ckanext.ytp_request.helper import get_safe_locale
+import ckanext.ytp_request.logic.action.create as ytp_request_action_create
+from ckanext.ytp_request.model import MemberRequest
+
+log = logging.getLogger(__name__)
+
+_SUBJECT_MEMBERSHIP_REQUEST = _("New membership request (%(organization)s)")
+_MESSAGE_MEMBERSHIP_REQUEST = _("""\
+    User %(user)s (%(email)s), %(job_title)s at %(affiliation)s, has requested membership to organization %(organization)s.
```
jonathansberry, line 20: Do we want this to be modern f-strings? Or is that a detail we are not so fussed about?

jonathansberry (changes requested):

ckanext/unaids/custom_user_profile/actions.py
tomeksabala: This PR required to duplicate a bit of code form ytp-request. I didn't want to focus on any other places then just where we're injecting custom unaids logic.

ckanext/unaids/custom_user_profile/actions.py
tomeksabala: I tried to explain this in the PR description. Custom user fields job_title and affiliation are defined in ckanext-unaids. yep-request shouldn't be aware of those.

ckanext/unaids/custom_user_profile/actions.py
tomeksabala: You can't use fstrings here. Those are string templates with enabled translations. They're concretized later in `mail_new_membership_request` with the use of old `%` instead of the more modern `.format()`.
But as I commented earlier I wanted to keep the amount of changes to a minimum.

jonathansberry (approved):

PR #310, opened 2023-10-17
ADX-854 Dataset locking

jonathansberry:
## Description

This PR enables sysadmins to "lock" a dataset, thereby making it read only to everyone in the system (including sysadmins). Locked datasets must first be "unlocked" by a sysadmin in order for further edits to take place. 

With this PR comes a minor improvement to the dataset edit form: the delete, transfer and lock actions have been placed inside a "danger zone" at the bottom of the page. 

To achieve all this the following approach is taken:

- A metadata field `locked` is added to datasets - see related PR in unaids_data_specifications.  This field is used to track which datasets are locked and make them searchable by their locked status. 
- A validator `read_only` is added to allow declaration of scheming fields (included the new "locked" field) to be "read only".  These fields can only be edited if a flag `bypass_read_only` is included in the context when making the update/patch action. 
- A helper `dataset_lockable` is included to determine if a dataset is of a type which is lockable.  This helper looks at whether the field "locked" has been defined in the dataset scheming schema.
- An action `dataset_lock` has been written.  This sets the value of locked to true, and creates an associate release of the dataset.
- An action `dataset_unlock` as been written.  This sets the value of locked to be false, and deletes any associated release.
- An auth check "dataset_lock" has been created that only passes if the user is a sysadmin.
- A chained auth check has been added for `package_update`.  This auth check is the heart of the feature.  It cannot be bypassed even if you are sysadmin.  The auth check blocks any update/patch actions upon a locked dataset.
- A blueprint `dataset_lock` has been created providing wrappers around actions `dataset_lock` and `dataset_unlock` for the purposes of the UI. 
- A JS module `transfer_dataset` has been added so that the transfer_dataset UX has been moved out of the dataset edit form into a modal dialog opened from the danger zone.  
- A super simple JS module `tooltips` has been created to enable bootstrap tool tips on specific elements. 
- Various template changes and CSS changes have been added to achieve the desired UX.

## Testing

Tests for the two actions, the two auth checks, the helper, and the validator, are included.

## Internationalisation and Localisation

All user-facing strings should be internationalised, and new po files generated. 

## Checklist

- [x] The Jira ticket for this issue has been updated to "Ready to Review" or equivalent.
- [x] I have developed these changes in discussion with the appropriate project manager.
- [x] My code follows the general Fjelltopp documentation (see Confluence).
- [x] I have made corresponding changes to the Fjelltopp documentation (see Confluence).
- [x] I have rebased this branch with master.
- [x] New dependency changes have been committed.
- [x] I have added automated tests that prove my fix is effective or that my feature works.
- [x] New and existing tests pass locally with my changes.
- [x] My changes generate no new warnings.
- [x] I have performed a self-review of my own code.
- [x] I have assigned at least one reviewer.


ckanext/unaids/actions.py
```diff
@@ -264,3 +264,61 @@ def package_create(next_action, context, data_dict):
...
+        context,
+        {"id": dataset_id, "locked": True}
+    )
+    try:
+        t.get_action("dataset_version_create")(context, {
+            "dataset_id": dataset_id,
+            "name": locked_name,
+            "notes": t._("This dataset has been made read-only by a system administrator")
+        })
+    except t.ValidationError:
```
tomeksabala: The check if dataset version with name `locked_name` can be created be better suited for the helper `h.dataset_lockable(dataset_id)`
In case this rollback happens you'll end up with polluted activity stream with:
1. "locked" = true
2. "locked" = false

ckanext/unaids/actions.py
```diff
@@ -264,3 +264,61 @@ def package_create(next_action, context, data_dict):
...
+    )
+    try:
+        versions = t.get_action("dataset_version_list")(
+            context,
+            {"dataset_id": data_dict["id"]}
+        )
+        assert versions, "No versions exist"
+        latest_version = versions[-1]
+        assert latest_version["name"] == locked_name, f"Expected latest version '{locked_name}'"
+        t.get_action("version_delete")(context, {"version_id": latest_version['id']})
```
tomeksabala: `assert` statements are removed and not executed in production (non local development code). IMO it is quite likely that in real life data in prod db you can end up with locked datasets where the latest version is not the one with `locked_name`.
You even allow to modification of such datasets with `context['bypass_read_only'] = True`

I suggest filtering `version_list` by latest occurrence of `name == locked_name` and removing this one.

ckanext/unaids/actions.py
```diff
@@ -264,3 +264,61 @@ def package_create(next_action, context, data_dict):
...
+    except Exception as e:
+        log.exception(f"Failed to delete version '{locked_name}'' "
```
tomeksabala, line 317: log.exception automatically adds whole stack trace in logs. No need to add `{e}` here.

ckanext/unaids/auth.py
```diff
@@ -1,12 +1,33 @@
...
+        'msg': 'You cannot carry out this action'
+    }
+
+
+@toolkit.chained_auth_function
+@toolkit.auth_sysadmins_check
+def package_update(next_auth_action, context, data_dict):
+    result = next_auth_action(context, data_dict)
+    locked = context['package'].extras.get("locked", 'false')
+    if locked in ['true', True]:
```
tomeksabala: !!!!! Try this:
```
>>> locked = 'True'
>>> locked in ['true', True]
False
```
Better to properly convert value to boolean on variable def:
```
locked = toolkit.asbool(context['package'].extras.get("locked", 'false'))
```

tomeksabala (changes requested):
[edit]: After reviewing the rest of changes I don't have anything more to add. It ended up as a big change with a well-thought design.

A partial review with a few comments I have so far. I'll add more later today.

ckanext/unaids/actions.py
jonathansberry: Yes I see the point you are making.  I think it doesn't fit in `dataset_lockable` perhaps because it serves a different purpose better articulated through a subtly different name e.g. "dataset_type_lockable" or some such name. But we could develop some separate function that checks the releases list prior to patching the dataset.

ckanext/unaids/actions.py
jonathansberry: That's an interesting bit of information thank you - I did not know the assert statements would not be executed. I think your logic here is sensible.

ckanext/unaids/actions.py
jonathansberry: Yes good catch I changed `log.error` to `log.exception` but did not change the log message.

ckanext/unaids/auth.py
jonathansberry: Yes I wasn't happy with this either. `toolkit.asbool` is what I needed. Thankyou!

jonathansberry:
Thanks for the review @tomeksabala I have addressed those points. 

I have also realised the read_only validator needed some updates to make the facet filtering work.  Unless an admin had locked a dataset, the value of `locked` was never initialised, so it doesn't show when filtering datasets by `locked`. Basically I realised we need to be able to initialise read_only fields to some default value.

Please feel free to address any issues you see need addressing whilst I am away if you want to get the PR merged quickly.

tomeksabala (changes requested):
Please address this issue when you're back and we can go ahead with deploying this change:
>I have also realised the read_only validator needed some updates to make the facet filtering work. Unless an admin had locked a dataset, the value of locked was never initialised, so it doesn't show when filtering datasets by locked. Basically I realised we need to be able to initialise read_only fields to some default value.

tomeksabala (approved):

PR #315, opened 2023-12-07
ADX-1058 Add required email & name field to the user edit form

jonathansberry:
[Sentry reports ](https://fjelltopp.sentry.io/issues/4521766833/?referrer=jira_integration)a key error every time someone tries to update their profile pic or description. 
```
KeyError: 'email'
  File "flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "ckan/config/middleware/../../views/user.py", line 287, in post
    email_changed = data_dict[u'email'] != g.userobj.email

'email'
```

The flask view for post requests sent to edit user details expects/requires an email field.  We can't re write the flask view easily, so the simpler solution is to ensure that the email field is included in the form sent to the view.  Upon making this change I realised that the form validation also expects/requires a name field.  

This PR then makes two simple changes - it adds the name and email fields as hidden inputs to the form. These fields will never actually be changed through the form, but must be supplied as part of the form in order for the CKAN view to function. 

Happy to hear if anyone thinks there is a better way to resolve the problem.

## Checklist

### DIFF
--- a/ckanext/unaids/theme/templates/user/edit_user_form.html
+++ b/ckanext/unaids/theme/templates/user/edit_user_form.html
@@ -45,7 +45,8 @@
 
   </fieldset>
 
-   <div>
+   <div class="hidden">
+      <!-- Hidden inputs required for the CKAN flask view to work -->
       <input type="hidden" name="old_password"/>
       <input type="hidden" name="password1"/>
       <input type="hidden" name="password2"/>
--- a/ckanext/unaids/theme/templates/user/edit_user_form.html
+++ b/ckanext/unaids/theme/templates/user/edit_user_form.html
@@ -17,7 +17,7 @@
                   <li>{{ data.email }}</li>
                   <li>&nbsp;</li>
                   <li>{% if data.job_title %}{{ data.job_title }}{% else %} {{_('No job title specified')}} {% endif %}</li>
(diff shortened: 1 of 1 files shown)


ckanext/unaids/theme/templates/user/edit_user_form.html
```diff
@@ -49,6 +49,8 @@
       <input type="hidden" name="old_password"/>
       <input type="hidden" name="password1"/>
       <input type="hidden" name="password2"/>
+      <input type="hidden" name="email" value="{{data.email}}" />
+      <input type="hidden" name="name" value="{{data.name}}" />
```
tomeksabala, line 54: Ok, so look like this whole `<div>` only purpose is to fulfill original ckan user controller needs after our custom changes to the user edit form: 1. removing the password fields and 2. Makin user details read only

I think it makes most sense to make the whole div hidden and also add a comment explaining why we need to dummy container with a few form fields not visible to the user.

tomeksabala (changes requested):

tomeksabala (approved):

PR #316, opened 2023-12-11
ADX-1046 Stop metadata changes triggering validation

jonathansberry:
## Description

This PR employs exactly the same technique/approach used in ckanext-validation to check whether a resource_update request is updating the actual file or just the metadata., and then to trigger validation appropriately. Here we are just triggering the "validate whole package" action. 

Whether or not a resource_update request includes an update to the underlying file cannot be easily checked in the after_update hook, so we have to check before_update, keep a record of which resources are having their file updated, and then trigger validation accordingly from that record in the after_update hook. This is how ckanext-validation does it. 

The result is that a metadata change to the geo-data does not trigger validation of the whole package. 

A template change is also included to ensure that metadata persists across updates made through the UI. 

## Testing

Regression test included as well as a test to check the validate_packge metadata element works. 

## Checklist

### DIFF
--- a/ckanext/unaids/plugin.py
+++ b/ckanext/unaids/plugin.py
@@ -71,7 +71,6 @@ class UNAIDSPlugin(p.SingletonPlugin, DefaultTranslation):
     This plugin implements the configurations needed for AIDS data exchange
 
     """
-    resources_to_validate_package = {}
     p.implements(p.IClick)
     p.implements(p.IConfigurer)
     p.implements(p.IFacets, inherit=True)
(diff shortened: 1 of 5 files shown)


ckanext/unaids/plugin.py
```diff
@@ -71,7 +71,7 @@ class UNAIDSPlugin(p.SingletonPlugin, DefaultTranslation):
     This plugin implements the configurations needed for AIDS data exchange
 
     """
-
+    resources_to_validate_package = {}
```
tomeksabala: This looks broken. Are you sure that it safe to use this dict like that? `UNAIDSPlugin` is a singleton plugin. I feel like `resources_to_validate_package` is going to be shared across all calls to CKAN after it's initialized once during startup.

I've seen your comment and took a look at the implementation in ckanext-validation. Let's hope that it's going to work as advertised ;)
Can you move this declaration to below `# IResourceController` so it's clearer on where this things is needed and used?

ckanext/unaids/plugin.py
```diff
@@ -238,6 +238,11 @@ def before_create(self, context, resource):
         return self._process_schema_fields(resource)
 
     def before_update(self, context, current, resource):
+        file_uploaded = current.get("url") != resource.get("url")
+
+        if file_uploaded and resource.get("validate_package"):
+            self.resources_to_validate_package[resource['id']] = True
+
```
tomeksabala: This block should be inside `if _data_dict_is_resource()` as in CKAN 2.9 the same `before_update` is still called for dataset & resource actions.

ckanext/unaids/plugin.py
```diff
@@ -208,8 +207,9 @@ def after_update(self, context, pkg_dict):
                     dataset_id=pkg_dict["id"],
                     recipient_org_id=org_to_allow_transfer_to[0],
                 )
-        if pkg_dict.get("validate_package") and not context.get("_dont_validate"):
-            logging.warning("VALIDATING ENTIRE PACKAGE")
+
+        if pkg_dict['id'] in self.resources_to_validate_package and not context.get("_dont_validate"):
```
tomeksabala: If user passes `context["_dont_validate"]` we skip the whole dataset validation but shouldn't you clear `self.resources_to_validate_package[pkg_dict['id']]` so that the validation is skipped permanently (and not repeated next time dataset is updated)?

ckanext/unaids/plugin.py
```diff
@@ -208,8 +207,9 @@ def after_update(self, context, pkg_dict):
...
+            del self.resources_to_validate_package[pkg_dict['id']]
```
tomeksabala: `self.resources_to_validate_package` is somethimes used with `resource[id]` and sometimes with `dataset[id]`. Could you explain how exactly `self.resources_to_validate_package` works?

tomeksabala (changes requested):

ckanext/unaids/plugin.py
jonathansberry: As you have said in your other comment, in CKAN 2.9 pkg_dict here might be a resource or a package.  What is missleading is the pkg_dict variable name.  I'll change it to data_dict.

ckanext/unaids/plugin.py
jonathansberry: Yes, whilst not needed for functionality, it will definitely improve readability and future proof it against the possibility of someone adding a url field to the dataset schema. I like this.

ckanext/unaids/plugin.py
jonathansberry: Yes the logic looks a little funny.  Upon further investigation I have realised we can ditch the _dont_validate context param entirely. This was a hack to avoid package_validation being triggered by updating the validation_status metadata element after validation.

tomeksabala (approved):

PR #324, opened 2026-05-18
chore (1/8): update CI workflow and deps for CKAN 2.11 + Python 3.10

A-Souhei:
## Summary

- Replace CKAN 2.9/2.10 matrix with a single CKAN 2.11 + Python 3.10 job
- Bump Docker images to `ckan-dev:2.11-py3.10`, `ckan-solr:2.11-solr9`, `ckan-postgres-dev:2.11`
- Add `safe.directory` git config step required inside the CKAN 2.11 container
- Update `actions/checkout`, `setup-python`, `setup-node` to `@v6`
- Pin extension deps to CKAN-2.11-compatible branches/commits; inline `collections.abc` patch for `ckanext-authz-service` (Python 3.10 compat)
- Add `numpy>=1.21.0,<2.0`, `frictionless>=5.0.0`, `tableschema>=1.21.0` to `requirements.txt`
- Fix `test.ini` core config path for the CKAN 2.11 container layout (`/srv/app/src/ckan/test-core.ini`)

## Notes

- This is PR #1 of a CKAN 2.11 migration sequence — CI/workflow and dependency changes only. No application code is touched here; subsequent PRs will handle runtime and test-suite fixes.
- The tests will definitively fail.

The `test.ini` path set here (`/srv/app/src/ckan/test-core.ini`) is also overwritten at runtime by the `sed` line in the workflow, so both paths end up consistent in CI.

## Test plan

- [ ] CI workflow runs on this PR and uses the correct CKAN 2.11 images
- [ ] Lint job passes on Python 3.10

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,16 +56,16 @@ jobs:
       run: |
         pip install -r requirements.txt
         pip install -r dev-requirements.txt
-        pip install -e "git+https://github.com/fjelltopp/ckanext-scheming@fix/ckan-2.11-python-3.10-compatibility#egg=ckanext-scheming"
-        pip install -e "git+https://github.com/fjelltopp/ckanext-validation@toavina/update-python#egg=ckanext-validation"
-        pip install -e "git+https://github.com/fjelltopp/ckanext-ytp-request@toavina/update-python#egg=ckanext-ytp-request"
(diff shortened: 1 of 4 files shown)


ChasNelson1990 (approved):
Yup, looks good to me

PR #325, opened 2026-05-18
chore (2/8): make plugin load and import under CKAN 2.11

A-Souhei:
## Summary

Minimal runtime fixes to get `ckanext-unaids` importing under CKAN 2.11 + Python 3.10, with all 14 test modules collecting cleanly. Test fixture and plugin-list work is deferred to the next PR.

### Plugin / framework compat
- Flask 2.2+: replace `_request_ctx_stack.top.current_user` with `g.current_user`; drop dead `repoze.who` / `zope.interface` / `six` imports.
- Drop CKAN-2.11-removed `before_request` registration and `_identify_user_default` shim — the framework's middleware handles user identification now.
- Rewrite `UNAIDSReclineView` as a `p.SingletonPlugin` + `IResourceView` using DataTables, since `ckanext.reclineview` was removed in 2.11.

### SQLAlchemy 2.0
- `dataset_transfer/model.py`: replace removed `Table.exists()` with `inspect(meta.engine).get_table_names()`; bind `create()` to `meta.engine` with `checkfirst=True`.

### Action / controller hooks
- `package_activity_list`: wrap the release-name enrichment in a defensive `try/except` so it stays a no-op when the versions plugin is unavailable.
- Rename `IResourceController` hooks for 2.11 (`before_create/update/show` → `before_resource_create/update/show`).
- Move the `validate_package` trigger into a new `after_resource_update`, which also fixes a pre-existing bug where `after_update` (IPackageController) was dereferencing `package_id` on a resource dict.

### Userobj / context safety
- Guard `toolkit.g.userobj` reads in `identify()`, `after_saml2_login`, and `auth.package_update` (may be unset during early request lifecycle); fall back to `model.Package.get` when context lacks `package`.

### Dataset transfer
- Switch the org-move `package_update` call to use the site user instead of an empty user — needed because 2.11's activity plugin requires a real user in context, and the org-move validation runs even with `ignore_auth=True`.

### Test imports
- Replace `nose` imports with stdlib `assert` shims; switch the standalone `mock` package for `unittest.mock`.

## Stacking note

This PR targets `chore/ckan-211-pr1-ci-workflow` (PR #324). It should be rebased onto `development` after #324 merges.

## Test plan

- [ ] Plugin imports cleanly under CKAN 2.11
- [ ] All 14 test modules collect (no `ImportError` at collection)
- [ ] Tests that don't depend on activity-plugin / scheming fixtures pass
- [ ] Remaining setup errors are limited to the `package_activity_list` chained-action lookup (handled in the follow-up test-suite PR)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,16 +56,16 @@ jobs:
       run: |
         pip install -r requirements.txt
         pip install -r dev-requirements.txt
-        pip install -e "git+https://github.com/fjelltopp/ckanext-scheming@fix/ckan-2.11-python-3.10-compatibility#egg=ckanext-scheming"
-        pip install -e "git+https://github.com/fjelltopp/ckanext-validation@toavina/update-python#egg=ckanext-validation"
-        pip install -e "git+https://github.com/fjelltopp/ckanext-ytp-request@toavina/update-python#egg=ckanext-ytp-request"
+        pip install -e "git+https://github.com/fjelltopp/ckanext-scheming@development#egg=ckanext-scheming"
+        pip install -e "git+https://github.com/fjelltopp/ckanext-validation@development#egg=ckanext-validation"
+        pip install -e "git+https://github.com/fjelltopp/ckanext-ytp-request@chas/update-python#egg=ckanext-ytp-request"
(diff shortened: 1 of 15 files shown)


ChasNelson1990 (dismissed):
LGTM

ChasNelson1990 (approved):
LGTM

PR #326, opened 2026-05-18
chore (3/8): align test suite with CKAN 2.11 plugin/route changes

A-Souhei:
## Summary

**Single cause: making the test suite run under CKAN 2.11.** Three structural changes, each tied to a concrete 2.11 change — no test logic was rewritten here (the per-test body fixes are split into #327). Two templates are touched; both are activity-plugin route renames (same cause as #330), needed because the tests render those pages.

This PR is the "make the suite executable" layer; #327 then fixes the individual test bodies. Split this way per the client's request to keep each PR to a single concern.

### 1. `conftest.py` — `clean_db_with_migrations` fixture *(cause: 2.11 `clean_db` rebuilds core schema only)*
A 7-line fixture calling CKAN's own `migrate_db_for("activity")` and `migrate_db_for("pages")` to re-apply the upstream Alembic migrations `clean_db` doesn't include in 2.11:
- activity's `permission_labels text[]` column (activity is decoupled into a plugin in 2.11)
- `ckanext-pages` tables (third-party plugin)

It no-ops for tests that don't load each plugin. **This deliberately replaces the source branch's raw `ALTER TABLE` / `CREATE TABLE` SQL with CKAN's canonical `migrate_db_for` pattern** — directly addressing the earlier review note about hand-written SQL. **Required:** without it, tests touching activity/pages fail on missing column/tables.

### 2. Per-test-class `ckan.plugins` markers *(cause: activity decoupled to a plugin in 2.11)*
Prepend `activity` to every `ckan_config('ckan.plugins', ...)` marker that lists `unaids`; add `ytp_request` to `test_helpers.py`. Our `unaids` plugin chains `package_activity_list`, which fails to register unless `activity` is loaded first (and `member_request_create` needs `ytp_request`). **Required:** without it, loading `unaids` in those tests raises "action not found for chained action."

### 3. Two template route renames *(cause: activity-plugin route renames in 2.11 — same as #330)*
`package/read_base.html`, `package/releases/list.html`:
- `'dataset.activity'` → `'activity.package_activity'` (endpoint moved into the activity blueprint)
- `request.args.activity_id` → `request.view_args.activity_id` (2.11 moved activity_id from query string to a URL-path variable — verified working)

**Required:** these pages are rendered by the suite; the old route 404s under 2.11.

## Required?

Yes — every change maps to a specific 2.11 behaviour change (schema-rebuild scope, activity-as-plugin, activity-route renames). Nothing cosmetic or speculative.

## Test progression

| Stage | Passed | Failed | Errors |
|---|---|---|---|
| After #325 | 24 | 3 | 134 |
| After this PR (#326) | 148 | 11 | 2 |
| After #327 (per-test bodies) | **160** | 0 | 0 |

The 11 remaining failures after this PR are the per-test-body issues, fixed in **#327** (not pending follow-up — that PR is already open and green).

## Stacking note

Targets `chore/ckan-211-pr2-runtime-fixes` (#325). Merge order: #324 → #325 → this → #327 → #328 → #329 → #330.

## Test plan

- [x] Full suite green at the top of the stack (160 passed / 1 skipped)
- [ ] CI green on this branch

### DIFF
--- a/ckanext/unaids/tests/conftest.py
+++ b/ckanext/unaids/tests/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+import ckan.plugins as p
 
 from ckan.tests import factories
 from ckanext.unaids.tests import unaids_db_setup, create_version
@@ -7,8 +8,19 @@ from ckanext.validation.model import tables_exist, create_tables
 from ckanext.ytp_request.tests import ytp_request_db_setup
 
 
+@pytest.fixture
+def clean_db_with_migrations(clean_db, with_plugins, migrate_db_for):
+    # CKAN 2.11: clean_db rebuilds the core schema only. Plugin-owned tables
+    # (activity.permission_labels as text[], ckanext_pages) need their Alembic
+    # migrations re-applied. Only migrate plugins actually loaded for this test
+    # so test classes don't need to declare plugins they don't use.
+    for plugin_name in ("activity", "pages"):
+        if p.plugin_loaded(plugin_name):
+            migrate_db_for(plugin_name)
+
+
 @pytest.fixture(autouse=True)
-def unaids_setup(clean_db):
+def unaids_setup(clean_db_with_migrations):
     if not tables_exist():
         create_tables()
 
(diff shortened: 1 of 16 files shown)


ChasNelson1990 (dismissed):
LGTM

PR #327, opened 2026-05-18
chore (4/8): fix per-test bodies for CKAN 2.11 behavior changes

A-Souhei:
## Summary

Last batch of CKAN 2.11 test fixes, deferred from #326 so that PR could stay focused on structural changes (fixture + markers + route renames). Each fix here is a real CKAN 2.11 behavior change needing a small per-test edit.

## Changes per file

- **`test_plugin.py`** — pass a Sysadmin user + context to `resource_create` calls; CKAN 2.11 records the creator on resources and now errors with `user_id: User not found` when context has no user. Affects the format-guess test, the validator-was-called test, and the `validate_package_resource` fixture used by `TestValidatePackage`.
- **`test_actions.TestPackageCreate`** — same `user_id` requirement, plus create the org with the same Sysadmin so `package_create`'s `owner_org` membership check is satisfied at the model layer (mandatory in 2.11).
- **`test_actions_dataset_lock.TestDatasetUnlock`** — thread the `locked_dataset` fixture's user through `dataset_unlock` and `version_delete`. The activity plugin's `package_changed` signal now requires a real user in context to record an activity row.
- **`test_auth_logic`** — collapse the two HTTP regression tests (`test_using_api_key`, `test_using_api_token`) into a single `call_action`-based test (per the audit's f141b1d). API key auth is gone in 2.11; API-token POSTs via the test client are subject to CSRF here, and the `call_action` variant exercises the same auth path this regression class was designed to guard.
- **`test_blueprints.TestMemberLists`** — compare against `org_admin` / `org_editor` / `org_member` fixture emails instead of the hardcoded `@ckan.org` strings. `factories.User()` emits non-deterministic emails in 2.11.
- **`test_giftless_backend`** — drop `from six import StringIO` for `from io import BytesIO`. Add `with_plugins` + user context to `TestResourceUrlEncoding`.

## Test results

| Stage | Passed | Failed | Errors |
|---|---|---|---|
| Before #324 | 0 | — | 161 collection |
| After #325 | 24 | 3 | 134 |
| After #326 | 148 | 11 | 2 |
| **After this PR** | **160** | **0** | **0** |

(1 test still skipped — the pre-existing giftless-authz blocker tracked at `datopian/ckanext-authz-service#24`.)

## Stacking note

Targets `chore/ckan-211-pr3-test-suite` (#326). To merge order: #324 → #325 → #326 → this.

## Test plan

- [ ] CI green on this branch
- [ ] After rebase onto `development` (after #324/#325/#326 land), full test suite passes

### DIFF
--- a/ckanext/unaids/tests/test_actions.py
+++ b/ckanext/unaids/tests/test_actions.py
@@ -173,9 +173,11 @@ class TestPackageCreate():
             )
 
     def test_create_dataset_without_type_creates_one_with_default_type_of_dataset(self):
-        organization = factories.Organization()
+        user = factories.Sysadmin()
+        organization = factories.Organization(user=user)
         dataset = call_action(
             'package_create',
+            {'user': user['name']},
             name="some-name",
             owner_org=organization['name'],
             title="Dataset without type"
@@ -184,9 +186,11 @@ class TestPackageCreate():
         assert dataset["type"] == "dataset"
 
     def test_create_dataset_with_valid_type(self):
-        organization = factories.Organization()
+        user = factories.Sysadmin()
+        organization = factories.Organization(user=user)
         dataset = call_action(
             'package_create',
+            {'user': user['name']},
             name="some-name",
             type="test-schema",
             title="Dataset with valid type",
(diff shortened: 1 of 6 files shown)


ChasNelson1990 (dismissed):
LGTM

ChasNelson1990 (approved):
LGTM

PR #328, opened 2026-05-18
chore (5/8): migrate templates from Bootstrap 3 to Bootstrap 5

A-Souhei:
## Summary

**Single cause: CKAN 2.11 ships Bootstrap 5 (CKAN 2.9 was Bootstrap 3).** Every change in this PR is a mechanical Bootstrap 3 → 5 class/attribute rename required for our existing templates to keep rendering correctly. **There is no visual redesign, no new layouts, and no colour or spacing changes** — only renames of classes/attributes that Bootstrap 5 removed or renamed.

This directly answers the client's question *"are these all to do with the bootstrap version changing? If so, that could be 1 PR"* — yes, and this is that one PR. Changes for other reasons (CSRF, activity-plugin routes) are split into their own PRs (#329, #330).

> Note: the large "new templating" from the original branch (the icon double-encoding rewrite — ~18 template overrides + a custom `helpers.py`) is **deliberately not here**. It was investigated and dropped: the underlying bug does not reproduce on CKAN 2.11.4, so those overrides are unnecessary.

## What changed (all BS3 → BS5 renames)

- `pull-right` / `pull-left` → `float-end` / `float-start`
- `data-toggle` / `data-target` → `data-bs-toggle` / `data-bs-target` (dropdowns, modal, collapse, tooltip)
- `data-placement` → `data-bs-placement`
- `text-right` → `text-end`
- `hidden` → `d-none` (incl. the Download All button + its `download_all.js` toggle, so show/hide works again)
- `col-xs-*` → `col-*`, `col-lg-offset-2` → `offset-lg-2`, `order-xs-2` → `order-2`, `hidden-xs hidden-sm` → `d-none d-md-block`
- `label label-info` / `label-danger` → `badge bg-info` / `bg-danger`
- format-chip CSS: added parallel `.badge[data-format]` selectors alongside the existing `.label[data-format]` ones so the colour coding still applies under BS5 markup (no colour values changed)

Plus one CKAN-2.11 (not Bootstrap) item that belongs with the homepage template work:
- New `home/index.html` shim that includes our `home/layout1.html`. CKAN 2.11 removed the `ckan.homepage_style` config that previously loaded our custom homepage; without this shim the homepage silently falls back to core CKAN's.

## Completeness

A repo-wide sweep was run for leftover BS3 classes (`pull-*`, `col-xs-*`, `hidden-*`, `data-toggle`, `data-placement`, `text-right`, `.hidden`, `label-*`). All functional leftovers are converted. Remaining `.label[data-format]` selectors are intentional (kept alongside the new `.badge[data-format]` for backwards compatibility); a `.col-xs-*` selector list in `custom.css` is a harmless padding rule (its `col-sm/md/lg` selectors still apply under BS5).

## Test plan

- [x] Full test suite green (160 passed / 1 skipped)
- [ ] Manual browser smoke: dropdowns open, dataset-transfer modal renders, format chips show in colour, Download All button hides/shows correctly, homepage shows the custom layout

### DIFF
--- a/ckanext/unaids/assets/css/ckanext-geoview.css
+++ b/ckanext/unaids/assets/css/ckanext-geoview.css
@@ -1,6 +1,8 @@
-.label[data-format=shp]{
+.label[data-format=shp],
+.badge[data-format=shp]{
     background-color: #91318c;
 }
-.label[data-format=geojson] {
+.label[data-format=geojson],
+.badge[data-format=geojson] {
     background-color: #8d2932;
 }
--- a/ckanext/unaids/assets/custom.css
+++ b/ckanext/unaids/assets/custom.css
@@ -134,45 +134,67 @@ a.dropdown-item{
 }
 
 .label[data-format=html],
-.label[data-format*=html] {
+.badge[data-format=html],
+.label[data-format*=html],
+.badge[data-format*=html] {
   background-color: #55a1ce;
 }
-.label[data-format=json]{
+.label[data-format=json],
+.badge[data-format=json]{
   background-color: #ef7100;
 }
 .label[data-format=xml],
-.label[data-format*=xml] {
+.badge[data-format=xml],
+.label[data-format*=xml],
+.badge[data-format*=xml] {
   background-color: #ef7100;
 }
 .label[data-format=text],
-.label[data-format*=text] {
+.badge[data-format=text],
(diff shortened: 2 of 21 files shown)


ChasNelson1990 (approved):
LGTM

PR #329, opened 2026-05-19
chore (6/8): add CSRF tokens for CKAN 2.11 forms and React file upload

A-Souhei:
## Summary

**Single cause: CKAN 2.11 enforces CSRF protection on POST forms.** CKAN 2.9 did not require extension templates to include a CSRF token; CKAN 2.11 rejects POST submissions without one. This PR adds the token to the forms this extension defines, and to the one React request that posts to CKAN.

This is split out from the Bootstrap work (#328) and the activity-route work (#326/#330) per the client's request to keep each PR to a single concern.

## What changed

Server-rendered forms — add `{{ h.csrf_input() }}`:
- `organization/bulk_process.html` (bulk-edit form)
- `package/collaborators/collaborator_new.html` (add/edit collaborator)
- `package/releases/base.html` (release create/edit/delete/restore)
- `user/edit_user_form.html` (user edit)
- `user/new_user_form.html` (user registration)

React file uploader:
- `FileInputComponent/src/FileUploader.js` — read the `_csrf_token` meta tag and send it as the `X-CSRFToken` header on the `/api/3/action/authz_authorize` POST (otherwise CKAN 2.11 rejects the upload-authorisation step).

## Required?

Yes — without these, the affected forms and the file-upload authorisation fail with a CSRF rejection on CKAN 2.11. No other changes are bundled in.

## Test plan

- [x] Full test suite green (160 passed / 1 skipped)
- [ ] Manual smoke: submit each of the 5 forms (expect success/redirect, not a 403 CSRF error); start a resource upload (expect `authz_authorize` to succeed)

PR #330, opened 2026-05-19
chore (7/8): fix organization nav routes and group_dict accessor for CKAN 2.11

A-Souhei:
## Summary

**Single cause: CKAN 2.11 moved the activity stream into its own plugin and renamed its routes; it also deprecated the Pylons `c` global.** This PR updates `organization/read_base.html` so the org page's nav renders again under CKAN 2.11.

Split from the Bootstrap work (#328) and CSRF work (#329) per the client's request to keep each PR to a single concern. (The equivalent dataset-page route renames live in #326, alongside the test-suite changes that needed them.)

## What changed (`organization/read_base.html`)

- Route names: underscore-style → Flask blueprint dot-style — `organization_read` → `organization.read`, `organization_about` → `organization.about`.
- Activity route: `organization_activity` → `activity.organization_activity` (the route moved into the activity blueprint in 2.11).
- Deprecated accessor: `c.group_dict` → `group_dict` (CKAN 2.11 passes `group_dict` to the template; the Pylons `c` global is deprecated).

## Required?

Yes — under CKAN 2.11 the old route names no longer resolve (the nav links would error), and the activity route only exists under the activity blueprint now.

## Scope note

Intentionally limited to this one template. Other templates that still reference `c.group_dict` (e.g. `organization/member_new.html`) are left as-is — they keep working via CKAN's deprecated-`c` compatibility shim and did not otherwise need touching. We only modernised `c.group_dict` where we already had to fix the routes, to keep the change need-driven.

## Test plan

- [x] Full test suite green (160 passed / 1 skipped)
- [ ] Manual smoke: visit /organization/<name> — nav shows Datasets / DHIS2 Sources / Activity Stream / About and each link resolves

### DIFF
--- a/ckanext/unaids/theme/templates/organization/read_base.html
+++ b/ckanext/unaids/theme/templates/organization/read_base.html
@@ -1,13 +1,13 @@
 {% ckan_extends %}
 
 {% block content_primary_nav %}
-  {{ h.build_nav_icon(group_type + '_read', _('Datasets'), id=c.group_dict.name) }}
+  {{ h.build_nav_icon(group_type + '.read', _('Datasets'), id=group_dict.name) }}
   <li>
-    <a  href="{{ h.url_for('/harvest', organization=c.group_dict.name) }}">
+    <a href="{{ h.url_for('/harvest', organization=group_dict.name) }}">
       <i class="fa fa-database"></i>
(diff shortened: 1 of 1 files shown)


ChasNelson1990 (approved):
LGTM

PR #331, opened 2026-05-21
chore (8/8): CKAN 2.11 post-migration cleanup & regression fixes

A-Souhei:
## Summary
Final PR in the CKAN 2.11 migration stack. Removes migration-orphaned dead code and fixes regressions surfaced during review/QA on the running deploy. Kept separate so the focused migration PRs stay clean.

> Hook/route claims below were verified against CKAN 2.11 source: `ckan/plugins/interfaces.py` (interface method names), `ckan/views/dashboard.py`, `ckanext/activity/views.py`.

## Dead code (migration leftovers — no behaviour change)
- **`auth_logic.py`** — remove `json_response_omitting_challenge_decider` (its only caller, the repoze.who `IChallengeDecider` registration, was removed in #325 when repoze.who left 2.11) and its test; remove the write-only `g.current_user = payload` (nothing reads it — CKAN's `current_user` is the separate flask-login proxy). The token-decode path is unchanged.
- **`svg_map_options.py` / `user_info_blueprint.py`** — drop the last `six` usages (`six.iteritems()` → `.items()`, `ensure_str()` → `str()`).
- **`user/edit_user_form.html`** — remove the dead "Regenerate API Key" button: CKAN 2.11 removed the `user.generate_apikey` route and `user_generate_apikey` auth function (API keys → API tokens), so the button raised `BuildError`.

## CKAN 2.11 regressions (renamed hooks/routes silently broke features)
CKAN 2.10/2.11 renamed the controller hooks (confirmed in `interfaces.py`): `IResourceController` → `before/after_resource_create|update|delete` (no `after_create`); `IPackageController` → `after_dataset_create|update|delete` (no `after_update`). Core never calls the old names, so the extension's old-named hooks were dead.

- **Routes** (`validate_user_affiliation.py`, `user/dashboard.html`) — `dashboard.index` → `activity.dashboard`. The dashboard news-feed view moved to the activity plugin in 2.11 (the `dashboard` blueprint now only has `/datasets`, `/groups`, `/organizations`; activity registers `/dashboard/`). The stale route name raised `BuildError`, breaking the post-login affiliation redirect and the dashboard nav tab.
- **Dataset-transfer emails** (`plugin.py`) — `after_update` → `after_dataset_update`, restoring the transfer-request notification email (silently dead since the rename). Reads `org_to_allow_transfer_to` top-level (scheming) with an `extras` fallback. Verified end-to-end (target org admins receive the email); covered by a new regression test.
- **Bulk-upload validation** (`plugin.py`, `logic.py`) — `after_create` → `after_resource_create`, which now triggers `resource_validation_run` for a stored resource with a schema + uploaded file + supported format. The bulk uploader creates resources with only `name`+`resource_type`; `schema`/`format` are applied as scheming defaults *after* ckanext-validation's `can_validate` gate runs, so validation was skipped — triggering it once the resource is stored fixes it. Separately, `validate_resource_upload_fields` no longer treats the standard CKAN `size` field as a giftless upload field (only `sha256`+`lfs_prefix` must be present-together and valid); the old `size` coupling made ckanext-validation's `validation_status` patch fail with *"Invalid blob storage upload fields"*, silently blocking validation on the create path.

## Bootstrap 5 theme regressions
- Restore the `.hidden` utility that BS5 dropped (un-hides autogenerated schema fields — title/url/tags/license/visibility — across all dataset forms); hide the validation "View Error Report" link for valid resources (`d-none`).
- Match production masthead nav link colour + hover/active styling.
- Remove the duplicate "Activity Stream" tab (the activity plugin already adds it via `super()`); stop listing every page in the top navbar (`build_pages_nav_main`).

## Tests
- Removed the test for the deleted `json_response_omitting_challenge_decider` (suite count −6 cases).
- Added `test_after_dataset_update_sends_transfer_email` — regression guard for the transfer hook (fails on the old `after_update`, passes on the fix).
- Fixed `test_send_dataset_transfer_emails_errors` — CKAN 2.11 factories auto-make the org creator an admin with an email, so the "no recipient" case was impossible; reworked to a genuine no-eligible-recipient case.
- Aligned `test_validate_resource_upload_fields` parametrized cases with the `size` decoupling.

## Verification
Full ckanext-unaids suite green (155 passed, 1 skipped); flake8 clean. Transfer-email and bulk-upload validation verified on the running deploy. Hook/route renames confirmed against CKAN 2.11 source.

## Stacking
Targets `chore/ckan-211-pr8-activity-routes` (#330) — tip of the migration chain.


ChasNelson1990 (approved):
LGTM - go! Go! Go!

PR #332, opened 2026-05-26
chore (4/8): fix per-test bodies for CKAN 2.11 behavior changes

A-Souhei:
## Summary
- Fixes test bodies to work under CKAN 2.11 stricter auth requirements (explicit sysadmin context for package/resource actions)
- Replaces `six.StringIO` with `io.BytesIO`
- Rewrites the OAuth2 regression test to use `call_action` with user context (HTTP-level auth unreliable in CKAN 2.11 test client)
- Fixes hardcoded email assertions to use dynamic factory values

Part of the CKAN 2.9 → 2.11 migration stack. Tests pass from this PR onwards.


A-Souhei:
I used this as a test, closing now

PR #337, opened 2026-07-09
Fix: OAuth2 JWKS-fetch errors should be 401, not unhandled 500

cooper667:
## Summary

Root cause of a live Naomi <-> ADR integration bug:
 SSO/Bearer-token API calls (e.g. `organization_list_for_user`) were returning a generic 500 Internal Server Error.

## Root cause

`ckanext.unaids.auth0_domain` and `ckanext.unaids.oauth2_api_audience` were not configured  With `AUTH0_DOMAIN` unset:

```python
jsonurl = urlopen("https://" + AUTH0_DOMAIN + "/.well-known/jwks.json")
```

raises an immediate `TypeError` (concatenating `str` + `None`), completely outside the `try/except` that only wrapped the later `jwt.decode()` call. Since `TypeError` isn't `OAuth2AuthenticationError`/`OAuth2AuthorizationError`, the registered Flask error handlers didn't catch it either, so it fell through to CKAN's generic unhandled-exception 500.

The missing config itself is fixed separately in `adx_develop` (deploy ini files). This PR fixes the underlying code fragility so the same class of issue (Auth0 outage, network blip, or any other future config mistake) degrades to a proper 401 instead of crashing.

## Fix

- Explicit early check: if `AUTH0_DOMAIN`/`API_AUDIENCE` aren't configured, raise `OAuth2AuthenticationError` (→ clean 401) instead of letting a `TypeError` propagate.
- Wrap the JWKS fetch (`urlopen` + `json.loads`) in `try/except`, logging the real exception and raising `OAuth2AuthenticationError` on any failure.

### DIFF
--- a/ckanext/unaids/auth_logic.py
+++ b/ckanext/unaids/auth_logic.py
@@ -6,7 +6,7 @@ from urllib.request import urlopen
 from flask import Response
 from jose import jwt
 
-from ckan.common import request, g
+from ckan.common import request, g, login_user
 from ckan.logic import ActionError
 from ckan.model import Session, User
 
@@ -40,6 +40,7 @@ def access_token_present_and_valid_and_user_authorized():
         user = find_user_by_saml_id(subject)
         g.userobj = user
         g.user = user.name
+        login_user(user)
 
         return True
 
--- a/ckanext/unaids/auth_logic.py
+++ b/ckanext/unaids/auth_logic.py
@@ -48,8 +48,17 @@ def access_token_present_and_valid_and_user_authorized():
 
 def validate_and_decode_token(encoded):
     token = extract_token(encoded)
-    jsonurl = urlopen("https://" + AUTH0_DOMAIN + "/.well-known/jwks.json")
-    jwks = json.loads(jsonurl.read())
+
+    if not AUTH0_DOMAIN or not API_AUDIENCE:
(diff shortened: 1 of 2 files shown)


ChasNelson1990 (approved):
LGTM

A-Souhei (approved):
Looks good

cooper667:
The IAuthenticator.identify() hook only set g.userobj/g.user directly,
but CKAN's API action dispatcher (ckan/views/api.py) builds its context
from Flask-Login's current_user, not g.user/g.userobj -- and
identify_user() unconditionally re-derives g.user/g.userobj from
current_user right after every IAuthenticator.identify() call, silently
overwriting what this plugin set. Net effect: Bearer-token requests
authenticated successfully (no error) but ran as the anonymous user,
so e.g. organization_list_for_user returned an empty list instead of
the caller's actual orgs.

This is a real regression from the CKAN 2.11 migration: the original
pre-2.11 code had a g.current_user = payload line, which the migration's
post-migration cleanup (https://github.com/fjelltopp/ckanext-unaids/pull/331) removed as 'confirmed write-only' -- true,
but the actual fix needed was calling Flask-Login's login_user(), the
same pattern ckanext-saml2auth already uses for CKAN 2.10+.

PR #343, opened 2026-07-16
Fix: membership requests invisible to org admins on CKAN 2.11 (Manage Members page)

A-Souhei:
image

## Problem

Since the CKAN 2.11 migration, organization admins have had **no clickable path to see or approve membership requests**.

CKAN 2.10 split the organization members UI into two pages (ckan/ckan@ce67c57337):

- `organization.members` → the public member list (old page, renders `members.html`)
- `organization.manage_members` → admin management (new page, renders `manage_members.html`) — **the Manage area's Members tab now links here**

Our `organization/members.html` override — which embeds the pending ckanext-ytp-request membership requests with Approve/Reject, plus affiliations, emails and the CSV download — was left targeting the old page, which no longer has any inbound link in the ADX UI. The requests table still existed, but only at a URL nobody is ever sent to.

## Change

Move the override to `organization/manage_members.html` — a pure `git mv`, no content change. The core template exposes the same `page_primary_action` and `primary_content_inner` blocks, so the override applies as-is. The old `organization.members` route falls back to ckanext-ytp-request's plainer override.

Note: core's manage_members page ships its own **CSV** button, which now sits beside our **Members' Emails** export (ours additionally includes emails and job titles). Harmless duplication; flagging for awareness.

## Related

Companion to fjelltopp/ckanext-ytp-request#39, which fixes the request-side bugs (Request to Join button, auth functions, request-listing query). Together they restore the full join → approve workflow on CKAN 2.11.

## Testing

Verified end-to-end on a local CKAN 2.11 stack through the real UI: org admin (non-sysadmin) → org page → Manage → Members tab → pending request visible with Approve/Reject/Modify → Approve succeeds → user appears in the member list.


ChasNelson1990 (approved):
LGTM

